PR #6, opened 2019-04-11
Update Dockerflie add build scripts and add Makefile

jharrington22:
This PR adds the build scripts required for deployment by the app-sre pipeline.


jharrington22:
@mwoodson I noticed there are no CRDs but that might be either because you don't need them or you are not finished development yet. Let me know an I can update this PR.

cc @jmelis

mwoodson:
There are no CRD's for this one, they aren't needed.

I looked over things here, and I assume these will work from the previous work you have done.

jharrington22:
@mwoodson this is ready for merging if you are happy!

PR #14, opened 2019-04-17
attempt to setup the clusterRole

mwoodson:


deploy/role.yaml
```diff
@@ -37,3 +37,28 @@ rules:
...
+  - clusterdeployments
+  - clusterdeployments/finalizers
+  - clusterdeployments/status
+  - syncsets
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
```
jewzaam: Why does this operator need to modify `ClusterDeployments`?

PR #15, opened 2019-04-17
only look at installed clusterdeployments

mwoodson:


mwoodson:
@jewzaam updated to include the proper test.

PR #21, opened 2019-05-20
Deploy needed objects for prometheus metrics when the operator comes up

fahlmant:
This PR ensures that when the operator is deployed, all the proper objects are placed so prometheus can scrape the custom metrics.

-The operator creates an deadmanssnitch-operator service to expose the metrics over on deploy
-The operator creates an deadmanssnitch-operator servicemonitor to tell prometheus how to digest the metrics
-Removes the operator-sdk metrics service for now
-Includes a prometheus-k8s role for the deadmanssnitch-operator namespace that has permissions to get, list and watch services, pods and endpoints.
-Includes a prometheus-k8s rolebinding to allow the prometheus-k8s serviceaccount from the openshift-monitoring namespace to use the aformentioned role.

When the operator and the role/rolebindings are deployed, the deadmanssnitch-operator namespace must be given the following label to get everything working: openshift.io/cluster-monitoring="true"

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,63 @@
+// Copyright 2019 RedHat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// UpdatePlaceholderGauge ...
+func UpdatePlaceholderGauge() {
+
+	metricPlaceholder.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(1))
```
robotmaxtron: > "name": "pagerduty-operator"
Looks like a typo, should be 'deadmanssnitch-operator'

fahlmant: Good catch :) Fixing.

jewzaam:
added lgtm from max's approval

PR #23, opened 2019-05-28
Exposing heartbeat metric, fixed a few typos

robotmaxtron:
Exposed heartbeat metric, corrected some typos in log output.


pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -171,10 +171,10 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 		for _, s := range snitches {
 			delStatus, err := r.dmsclient.Delete(s.Token)
 			if !delStatus || err != nil {
-				reqLogger.Info("Failed to delete the DMS from api.deadmanssnicth.com", "Namespace", request.Namespace, "Name", request.Name)
+				reqLogger.Info("Failed to delete the DMS from api.deadmanssnitch.com", "Namespace", request.Namespace, "Name", request.Name)
```
tparikh: "Namespace", request.Namespace, "Name", request.Name in the logging statement is redundant.

 reqLogger is created with values (reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)) in the start of reconcil loop and anything logged with reqLogger will always add the fields.

pkg/metrics/metrics.go
```diff
@@ -56,8 +57,33 @@ func RegisterMetrics() error {
...
+	if err != nil {
+		metricDeadMansSnitchHeartbeat.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(0))
+	}
+	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+		metricDeadMansSnitchHeartbeat.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(1))
+	} else {
+		metricDeadMansSnitchHeartbeat.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(0))
+	}
+
+	defer resp.Body.Close()
```
tparikh: defer statement should be close to where the connection is opened.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
tparikh: Above comment applies to various other logging statements in the file as well

pkg/metrics/metrics.go
```diff
@@ -56,8 +58,34 @@ func RegisterMetrics() error {
 	return nil
 }
 
-// UpdatePlaceholderGauge ...
-func UpdatePlaceholderGauge() {
+// UpdateMetrics updates all the metrics ever 5 minutes
+func UpdateMetrics() {
 
-	metricPlaceholder.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(1))
+	d := time.Tick(5 * time.Minute)
```
jewzaam: Want to be sure we're not going to hit rate limits, but can't find enough info on what those limits are.  There's mention of a "Rate Limited" response here https://deadmanssnitch.com/docs/api/v1 but I can't find anything more specific.  We have to keep in mind we have multiple clusters running this operator and can expect that to scale over time.

I recommend we find out more before we merge this in.

/hold

jewzaam:
Per DMS: "API rate limits are 1500 API requests per 5 minute fixed window"
Option to increase it if we hit it.  With this I don't expect we'll hit the limit with this change.

PR #31, opened 2019-07-22
using operator-custom-metrics library

ninataneja:
Replaces pkg/metrics/service.go with using github.com/openshift/operator-custom-metrics pkg to register and expose metrics.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,7 +9,8 @@ import (
 
 	"github.com/openshift/deadmanssnitch-operator/pkg/apis"
 	"github.com/openshift/deadmanssnitch-operator/pkg/controller"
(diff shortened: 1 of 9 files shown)


cmd/manager/main.go
```diff
@@ -29,8 +31,9 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost       = "0.0.0.0"
-	metricsPort int32 = 8383
+	metricsPath               = "/metrics"
+	metricsPort               = "8080"
+	secretWatcherScanInterval = time.Duration(10) * time.Minute
```
fahlmant: This can be removed (I think this got copied over from aws-account-operator on accident)

PR #32, opened 2019-08-12
Added initialize function to initialize a snitch 

rafael-azevedo:
This PR initializes a snitch after it has been created. Still requires local testing.


pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -240,6 +240,28 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 			}
 		}
 
+		// Get the snitch again to check status
+		snitches, err = r.dmsclient.FindSnitchesByName(snitchName)
```
c-e-brumm: Is there a reason to re-fetch the snitches? It's done just a few lines above and I don't see anything updating either snitches or snitchName between that line and this one.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -240,6 +240,28 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if len(snitches) > 0 {
```
c-e-brumm: This block and the block starting at 226 are executing on the same conditions. It might make the code easier to read if they where combined and their actions abstracted into descriptive function calls, doTheThing() doTheOtherThing(). A suggestion, not a a requirement.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
rafael-azevedo: We refetch the snitch because there is a possibility that a new snitch is created here . https://github.com/openshift/deadmanssnitch-operator/blob/190ec5c31df9b2e96a292ffdc2fa6eff884102c6/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go#L237 .

We want to get the state after the snitch is created to ensure the request has been completed by the API and we aren't just looking at the snitch returned by the create operation, but that it actually exists when pulling as well.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
rafael-azevedo: Here again, we are now checking the return of the second snitch fetch. We could of comibined all of this be fetching the snitch before that first if loop, but that would make the api calls everytime a reconcile loop is called. This limits the number of times we hit the API since it has to pass that first if statement here https://github.com/openshift/deadmanssnitch-operator/blob/190ec5c31df9b2e96a292ffdc2fa6eff884102c6/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go#L215

PR #36, opened 2019-09-13
add yaml validation

liamwazherealso:
As a SREP engineer I want to know that the CRDs for a project are valid yaml before deploying. A CRD with bad yaml will break a deploy and also one OLM bundle revision meaning that you cannot "roll forward" by just fixing them yaml and deploying again.

https://jira.coreos.com/browse/SREP-2101

### DIFF
--- a/hack/validate_yaml.py
+++ b/hack/validate_yaml.py
@@ -1,0 +1,42 @@
+# Usage
+# python validate_yaml.py path/to/file/or/dir
+
+import sys
+import yaml
+from os import listdir
+from os.path import isdir, isfile, join, splitext
+
+usage = 'usage: python validate_yaml.py path/to/file/or/dir'
+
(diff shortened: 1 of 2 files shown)


hack/validate_yaml.py
```diff
@@ -0,0 +1,42 @@
...
+for file_path in files:
+    _, ext = splitext(file_path)
+    if ext not in ['.yml', '.yaml']:
+        continue
+
+    print('validating {}'.format(file_path))
+    with open(file_path, 'r') as f:
+        data = f.read()
+    try:
+        yaml.safe_load(data)
```
rafael-azevedo: yaml.safe_load(data) should be yaml.safe_load_all(data) to support multiple documents in a yaml file.

liamwazherealso: updated

PR #41, opened 2019-10-04
Remove SyncSet if noalerts label is added

jewzaam:
https://jira.coreos.com/browse/SREP-1572

Refactored delete to a separate deleteDMS function and trigger it from both CD delete and the noalerts label being set.  It checks the finalizer first and bails if there is no finalizer so should be lightweight.

### DIFF
--- a/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
+++ b/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
@@ -143,8 +143,6 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	println("managed: " + instance.Labels[ClusterDeploymentManagedLabel])
-
 	// Just return if this is not a managed cluster OR has noalerts label set
 	if val, ok := instance.Labels[ClusterDeploymentManagedLabel]; ok {
 		if val != "true" {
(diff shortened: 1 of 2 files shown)


pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -142,6 +143,8 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	println("managed: " + instance.Labels[ClusterDeploymentManagedLabel])
```
wshearn: debug printing

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -154,10 +157,12 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, nil
 	}
 
-	// Return if alerts are disabled on the cluster
+	println("noalerts: " + instance.Labels[ClusterDeploymentNoalertsLabel])
```
wshearn: debug printing

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -330,3 +310,48 @@ func newSyncSet(namespace string, clusterDeploymentName string, snitchURL string
 	return newSS
 
 }
+
+func deleteDMS(r *ReconcileDeadMansSnitch, request reconcile.Request, instance *hivev1alpha1.ClusterDeployment, reqLogger logr.Logger) error {
+	println("deleteDMS: " + instance.Labels[ClusterDeploymentNoalertsLabel])
```
wshearn: debug printing

PR #43, opened 2019-10-07
Refactoring and additional tests

jewzaam:


pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -130,47 +126,9 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
...
-		reqLogger.Info("Not a managed cluster", "Namespace", request.Namespace, "Name", request.Name)
-		return reconcile.Result{}, nil
-	}
-
-	// Cleanup DMS then return if alerts are disabled on the cluster
-	if _, ok := instance.Labels[ClusterDeploymentNoalertsLabel]; ok {
-		reqLogger.Info("Managed cluster with Alerts disabled", "Namespace", request.Namespace, "Name", request.Name)
-		return reconcile.Result{}, deleteDMS(r, request, instance, reqLogger)
-	}
+	processCD, instance, err := utils.CheckClusterDeployment(request, r.client, reqLogger)
```
cblecker: err looks unhandled here

pkg/utils/utils.go
```diff
@@ -28,3 +38,88 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
...
+			return false, clusterDeployment, nil
+		}
+	} else {
+		// Managed tag is not present which implies it is not a managed cluster
+		reqLogger.Info("Is not a managed cluster")
+		return false, clusterDeployment, nil
+	}
+
+	// Return if alerts are disabled on the cluster
+	if _, ok := clusterDeployment.GetLabels()[config.ClusterDeploymentNoalertsLabel]; ok {
```
cblecker: If this label is set to "false", what do you expect to happen?

jewzaam: It will still skip things, not checking the value of the label at all.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
jewzaam: I have pushed an update to check and requeue.  The push is not showing up in the PR which is weird.  Hopefully it lands soon: https://github.com/jewzaam/deadmanssnitch-operator/commit/8f4d7ff04c2259951de6998fa184c39b3d6b116b

jewzaam: Squashed and force pushed.  Now requeues if there is any error.

pkg/controller/deadmanssnitch/deadmanssnitch_controller_test.go
```diff
@@ -164,8 +201,8 @@ func nonManagedClusterDeployment() *hivev1alpha1.ClusterDeployment {
 // return a ClusterDeployment with Label["noalerts"] == ""
 func noalertsManagedClusterDeployment() *hivev1alpha1.ClusterDeployment {
 	labelMap := map[string]string{
-		ClusterDeploymentManagedLabel:  "true",
-		ClusterDeploymentNoalertsLabel: "",
+		config.ClusterDeploymentManagedLabel:  "true",
```
jharrington22: Double spacing

PR #44, opened 2019-10-08
use pre-created secret instead of embed in syncset

bmeng:
As requested in https://jira.coreos.com/browse/SREP-1831
And discussed with @wanghaoran1988 , he suggested that we should do the clean up things manually on the running clusters.

### DIFF
--- a/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
+++ b/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
@@ -155,13 +155,44 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 	}
 
 	ssName := request.Name + config.SyncSetPostfix
+	refSecretName := request.Name + config.RefSecretPostfix
+
 	// Check to see if the SyncSet exists
 	err = r.client.Get(context.TODO(),
 		types.NamespacedName{Name: ssName, Namespace: request.Namespace},
 		&hivev1alpha1.SyncSet{})
+
 	if errors.IsNotFound(err) {
 		// create new DMS SyncSet
 		reqLogger.Info("SyncSet not found, Creating a new SynsSet", "Namespace", request.Namespace, "Name", request.Name)
+
+		newSS := newSyncSet(request.Namespace, refSecretName, request.Name)
+
+		// ensure the syncset gets cleaned up when the clusterdeployment is deleted
+		if err := controllerutil.SetControllerReference(instance, newSS, r.scheme); err != nil {
+			reqLogger.Error(err, "Error setting controller reference on syncset", "Namespace", request.Namespace, "Name", request.Name)
+			return reconcile.Result{}, err
+		}
+		if err := r.client.Create(context.TODO(), newSS); err != nil {
+			reqLogger.Error(err, "Error creating syncset", "Namespace", request.Namespace, "Name", request.Name)
+			return reconcile.Result{}, err
+		}
+		reqLogger.Info("Done creating a new SyncSet", "Namespace", request.Namespace, "Name", request.Name)
+
+	} else {
+		reqLogger.Info("SyncSet Already Present, nothing to do here...", "Namespace", request.Namespace, "Name", request.Name)
+		// return directly if the syscset already existed
+		return reconcile.Result{}, nil
+	}
+
+	// Check if the secret exists
+	err = r.client.Get(context.TODO(),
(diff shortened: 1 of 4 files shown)


bmeng:
@jewzaam 
I have added some test for the new changes, and also addressed the deletion part.
Please help review again.  Thanks.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -155,15 +155,47 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
...
+	err = r.client.Get(context.TODO(),
+		types.NamespacedName{Name: secretName, Namespace: request.Namespace},
+		&corev1.Secret{})
+
+	if errors.IsNotFound(err) {
+		// create new secret which will be referenced by SyncSet
+		reqLogger.Info("Secret not found, Creating a new Secret", "Namespace", request.Namespace, "Name", request.Name)
+
+		// snitchName := instance.Spec.ClusterName + "." + instance.Spec.BaseDomain
+		snitchName := "ds-test"
```
jewzaam:
```suggestion
		snitchName := instance.Spec.ClusterName + "." + instance.Spec.BaseDomain
```

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -244,21 +275,17 @@ func newSyncSet(namespace string, clusterDeploymentName string, snitchURL string
...
-							},
-							ObjectMeta: metav1.ObjectMeta{
-								Name:      "dms-secret",
-								Namespace: "openshift-monitoring",
-							},
-							Data: map[string][]byte{
-								config.KeySnitchURL: []byte(snitchURL),
-							},
+						Source: corev1.ObjectReference{
+							Name:      clusterDeploymentName + config.SecretPostfix,
```
jewzaam: Recommend passing in the source secret name as a function arg rather than reconstructing.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -292,6 +336,15 @@ func deleteDMS(r *ReconcileDeadMansSnitch, request reconcile.Request, instance *
 		reqLogger.Info("Deleted the DMS from api.deadmanssnitch.com", "Namespace", request.Namespace, "Name", request.Name)
 	}
 
+	// Delete the secret
+	reqLogger.Info("Deleting DMS secret", "Namespace", request.Namespace, "Name", request.Name)
+	err = utils.DeleteSecret(request.Name+config.SecretPostfix, request.Namespace, r.client, reqLogger)
```
jewzaam: Should delete the `Secret` after `SyncSet` simply based on the SS referencing the secret.

pkg/controller/deadmanssnitch/deadmanssnitch_controller_test.go
```diff
@@ -460,6 +474,26 @@ func verifyNoSyncSet(c client.Client, expected *SyncSetEntry) bool {
 	return true
 }
 
+func verifyNoSecret(c client.Client, expected *SecretEntry) bool {
```
jewzaam: Take a look again at testing `Secret` deletion.  I commented out this block that deleted the secret and all tests still passed.  I would have expected it to fail.  Not sure why at a glance.

https://github.com/openshift/deadmanssnitch-operator/pull/44/files#diff-cbcc4d0350863bcdb11e774f6cacbaf5R339-R345

bmeng:
The reason about why the `delete secret`. I assumed that the test "Test Deleting" should cover that, but actually it does not.

So I updated the test "Test Managed Cluster that later sets noalerts label", and now the deletion of secret should be covered.

And I also renamed some of the function name and arg name for the created secret to "refSecret", to avoid confusion.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
bmeng: forgot to change it back which is used for local testing...

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
bmeng: moved to function arg

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
bmeng: Changed the order

PR #67, opened 2020-07-27
Prometheus Metrics

arjunrn:
Following changes were made:

1. Removed existing metrics which was deadcode. It was not wired to anything and metrics were never collected. Also calling the API just to check if it's that exhausts the rate limits.
2. Added metric for operator reconcile duration time.
3. Added metric for API calls to the Dead Man's Snitch API
4. Added metric for API calls to the API server. Mostly copied from this [PR](https://github.com/openshift/aws-account-operator/pull/410) in [aws-account-operator](https://github.com/openshift/aws-account-operator/).

[Internal](https://issues.redhat.com/browse/OSD-4568) [Issues](https://issues.redhat.com/browse/OSD-4570)


cmd/manager/main.go
```diff
@@ -117,14 +117,9 @@ func main() {
...
-	// Setup all Controllers
-	if err := controller.AddToManager(mgr); err != nil {
-		log.Error(err, "")
-		os.Exit(1)
-	}
-
+	localmetrics.Collector = localmetrics.NewMetricsCollector()
 	metricsServer := metrics.NewBuilder().WithPort(metricsPort).WithPath(metricsPath).
-		WithCollectors(localmetrics.MetricsList).
+		WithCollector(localmetrics.Collector).
```
fahlmant, line 122: A Collector is a prometheus type that all different type of metrics implement. The object passed here has a collectors list within it, but `WithCollector` is just expecting one metric

arjunrn: Right, so I implemented the `Collector` interface. Is there a problem with this approach?

fahlmant: Would all future metrics, even non-reconcile or api related go in this collector?

2uasimojo: AFAICT this is the pattern aws-account-operator uses.

https://github.com/openshift/aws-account-operator/blob/4bb28f759e24fdc03ccf943c9ec772150a421cd4/cmd/manager/main.go#L136
https://github.com/openshift/aws-account-operator/blob/4bb28f759e24fdc03ccf943c9ec772150a421cd4/pkg/localmetrics/localmetrics.go#L37-L55

seems fine by me.

pkg/controller/deadmanssnitch/deadmanssnitch_controller.go
```diff
@@ -125,6 +126,11 @@ func (r *ReconcileDeadMansSnitch) Reconcile(request reconcile.Request) (reconcil
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling DeadMansSnitch")
 
+	start := time.Now()
+	defer func() {
+		localmetrics.Collector.ObserveReconcile(time.Since(start).Seconds())
```
2uasimojo, line 131: Other impls put a "Reconcile complete" log message in here that includes the duration. Probably not a must-have, but could be nice.

pkg/dmsclient/instrumented_client.go
```diff
@@ -0,0 +1,125 @@
+package dmsclient
```
2uasimojo, line 1: This file seems like a really heavy way to instrument the dms client. Why not simply wrap the guts of the `do` func? I can see it'll be a little bit tricky to get the `method` label in place for `FindSnitchesByName`, if we even care to distinguish that from `ListAll`, but it should be doable.

arjunrn: IMO it's the same amount of complexity. In the `do()` method the path and payload will have to be inspected to determine which method is being called. I didn't want to change a lot of the existing code and instead chose to wrap it.

cmd/manager/main.go
fahlmant: huh, I didn't see that in the refactor. AFAIK, that's a pattern for proxying metrics from other monitoring stores and exporters, not for instrumenting metrics. But whatever, if it works

pkg/dmsclient/instrumented_client.go
2uasimojo: It's not about complexity -- I agree it would be doing the same thing either way, as it should. My concern is the code repetition, which carries with it the usual maintainability issues. There's also no need for the uninstrumented client to exist anymore.

pkg/dmsclient/dmsclient.go
```diff
@@ -103,14 +108,20 @@ func (c *dmsClient) newRequest(method, path string, body interface{}) (*http.Req
 }
 
 func (c *dmsClient) do(req *http.Request) (*http.Response, error) {
+	start := time.Now()
```
2uasimojo, line 111: I like this implementation a lot better, thank you.

pkg/localmetrics/localmetrics.go
```diff
@@ -16,51 +16,187 @@ package localmetrics
...
+// RecordSnitchCallDuration records the time taken to make a call to the Dead Man Snitch API
+func (m *MetricsCollector) RecordSnitchCallDuration(duration time.Duration, u *url.URL, method string) {
+	operation := m.parseSnitchCall(u.Path, method)
+	m.snitchCallDuration.With(prometheus.Labels{snitchMethodLabel: operation}).Observe(duration.Seconds())
+}
 
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		MetricDeadMansSnitchHeartbeat.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(0))
+func (m *MetricsCollector) parseSnitchCall(path, method string) string {
```
2uasimojo, line 122: It would be a bit more future-proof to add the `operation` as a `string` parameter to the `do` func. That way the label lives in the func it belongs to. (The only weird one would be `FindSnitchesByName`, which calls `ListAll` instead of `do`, but you're not handling that here anyway.)

pkg/dmsclient/dmsclient.go
```diff
@@ -103,14 +108,20 @@ func (c *dmsClient) newRequest(method, path string, body interface{}) (*http.Req
...
+	defer func() {
+		c.metricsCollector.RecordSnitchCallDuration(time.Since(start), req.URL, req.Method)
```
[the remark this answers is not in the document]
arjunrn: I don't think the response status codes are very important. If there is an error counter and we see an increase we can investigate. Also I will add the note about the `_count`.

pkg/dmsclient/dmsclient.go
arjunrn: :+1:

pkg/dmsclient/dmsclient.go
```diff
@@ -147,7 +157,7 @@ func (c *dmsClient) List(snitchToken string) (Snitch, error) {
 		return snitch, err
 	}
 
-	resp, err := c.do(req)
+	resp, err := c.do(req, "describe")
```
2uasimojo, line 160: `"describe"` made me raise an eyebrow as it doesn't match the name of the func... but it's actually the name of the func that's questionable, since we're retrieving a single resource. +1

pkg/localmetrics/localmetrics.go
```diff
@@ -16,51 +16,167 @@ package localmetrics
...
 )
 
-// UpdateMetrics updates all the metrics ever 5 minutes
-func UpdateMetrics() {
+type MetricsCollector struct {
+	ReconcileDuration  prometheus.Histogram
+	apiCallDuration    *prometheus.HistogramVec
+	snitchCallErrors   prometheus.Counter
+	snitchCallDuration *prometheus.HistogramVec
+	collectors         []prometheus.Collector
```
2uasimojo, line 40: Why is this here? I don't see it being used (except by `AddCollector`, which we also don't need, see below).

pkg/localmetrics/localmetrics.go
```diff
@@ -16,51 +16,167 @@ package localmetrics
...
+
+// ObserveReconcile records the duration of the reconcile loop of the operator
+func (m *MetricsCollector) ObserveReconcile(seconds float64) {
+	m.ReconcileDuration.Observe(seconds)
+}
+
+// AddCollector adds a collector to existing metrics
+func (m *MetricsCollector) AddCollector(collector prometheus.Collector) {
+	m.collectors = append(m.collectors, collector)
+}
```
2uasimojo, line 114: Pretty sure we don't need this. It's unused, and `MetricsCollector` only needs to satisfy `prometheus.Collector` afaict, yes?

(If we need it in the future, we can always add it then. YAGNI.)

pkg/localmetrics/localmetrics.go
```diff
@@ -16,51 +16,167 @@ package localmetrics
...
+
+// RecordSnitchCallDuration records the time taken to make a call to the Dead Man Snitch API
+func (m *MetricsCollector) RecordSnitchCallDuration(duration time.Duration, operation string) {
```
2uasimojo, line 117: nit: It would be nice to make these more consistent.
- Right now the names of the methods are `Add`, `Observe`, and `Record`. Can we pick one?
- We've got a `duration float64`, a `seconds float64`, and a `duration time.Duration` which does `.Seconds()` within the func. Can we pick one?

pkg/localmetrics/localmetrics.go
```diff
@@ -16,51 +16,167 @@ package localmetrics
...
-	if err != nil {
-		MetricDeadMansSnitchHeartbeat.With(prometheus.Labels{"name": "deadmanssnitch-operator"}).Set(float64(0))
+func NewMetricsCollector() *MetricsCollector {
+	return &MetricsCollector{
+		ReconcileDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:        "dms_operator_reconcile_duration_seconds",
+			Help:        "The duration it takes to reconcile a ClusterDeployment",
+			ConstLabels: map[string]string{"name": operatorName},
+		}),
+		// apiCallDuration times API requests. Histogram also gives us a _count metric for free but also includes errors.
```
2uasimojo, line 70: Noticed after the fact (and I think this is my mistake, I think I flagged the wrong location in my previous comment): This isn't the right place for this comment, as apiCallDuration does *not* count errors. The comment needs to go next to `snitchCallDuration`.

I'm fixing this in the fup.

2uasimojo: Addressed in #69.

pkg/localmetrics/localmetrics.go
2uasimojo: Addressed in #69.

pkg/localmetrics/localmetrics.go
2uasimojo: Addressed in #69.

pkg/localmetrics/localmetrics.go
2uasimojo: Addressed in #69.

PR #70, opened 2020-07-28
Adding DeadMansSnitchIntegration CRD to deadmanssnitch operator

R-Lawton:
This PR is for the following changes:
 1. Adding a DeadMansSnitchIntegration CRD that replaces the clusterDeployment controller which handled the creation of just one DMS secret and syncset. The new DMSI allows more then one DeadMans snitch and the matching secret and syncset to be created per cluster.

2. Watches
* Create/Update/Delete of a DeadMansSnitchIntegration will trigger a reconcile request for that DeadMansSnitchIntegration.
* Create/Update/Delete of a ClusterDeployment will trigger a reconcile request for each DeadMansSnitchIntegration that matches that ClusterDeployment (using the spec.clusterDeploymentSelector label selector).
* Create/Update/Delete of a Secret or SyncSet owned by a ClusterDeployment will trigger a reconcile request for each DeadMansSnitchIntegration that matches that ClusterDeployment 
## functionality:
1. Create multiple snitches,secrets and syncset per matching cluster deployment(that have the right matching labels)
2. Delete snitches from deadmanssnitch.com secrets and syncset by either deleting the clusterdeployment of the snitches etc you want deleted or by deleting the DMSI CR of the snitches etc you wanted deleted


deploy/role.yaml
```diff
@@ -62,4 +61,16 @@ rules:
   resources:
   - routes
   verbs:
-  - '*' 
+  - '*'
+- apiGroups:
+  - deadmansnitch.openshift.io
```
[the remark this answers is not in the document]
R-Lawton: Cool did a search for the single s and replaced with the double s aswell as added the .managed to

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+	// DeadMansSnitchOperatorNamespace is the namespace where this operator will run
+	DeadMansSnitchOperatorNamespace string = "deadmanssnitch-operator"
+	// DeadMansSnitchAPISecretName is the secret Name where to fetch the DMS API Key
+	DeadMansSnitchAPISecretName string = "deadmanssnitch-api-key"
+	// DeadMansSnitchAPISecretKey is the secret where to fetch the DMS API Key
+	DeadMansSnitchAPISecretKey string = "deadmanssnitch-api-key"
+	//DeadMansSnitchTagKey for tests
+	DeadMansSnitchTagKey string = "testTag"
+	//DeadMansSnitchFinalizer for tests
+	DeadMansSnitchFinalizer string = "testfinalizer"
```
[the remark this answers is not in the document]
jewzaam: Yes, `DeadMansSnitchTagKey` and `DeadMansSnitchFinalizer` are misnamed and out of place, please move to the test.

pkg/utils/utils.go
```diff
@@ -111,7 +114,7 @@ func DeleteSyncSet(name string, namespace string, client client.Client, reqLogge
 
 	// Only delete the syncset, this is just cleanup of the synced secret.
 	// The ClusterDeployment controller manages deletion of the pagerduty serivce.
```
jewzaam, line 116:
```suggestion
	// The ClusterDeployment controller manages deletion of the deadmanssnitch serivce.
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
```diff
@@ -29,13 +32,16 @@ import (
...
-	testOtherSyncSetPostfix = "-something-else"
+	testDeadMansSnitchintegrationName = "testDeadMansSnitchIntegration"
+	testClusterName                   = "testClusterName"
+	testNamespace                     = "testNamespace"
+	testSnitchURL                     = "https://deadmanssnitch.com/12345"
+	testSnitchToken                   = "abcdefg"
+	testTag                           = "test"
+	testAPIKey                        = "abc123"
+	testOtherSyncSetPostfix           = "-something-else"
+	testsecretReferencesName          = "pd-secret"
```
jewzaam, line 43:
```suggestion
	testsecretReferencesName          = "dms-secret"
```

pkg/apis/deadmanssnitch/v1alpha1/deadmanssnitchintegration_types.go
```diff
@@ -0,0 +1,60 @@
...
+	ClusterDeploymentSelector metav1.LabelSelector `json:"clusterdeploymentSelector"`
+
+	//name and namespace in the target cluster where the secret is synced
+	TargetSecretRef corev1.SecretReference `json:"targetSecretRef"`
+
+	//Array of strings that are applied to the service created in DMS
+	Tags []string `json:"tags"`
+
+	//The postfix to append to any snitches managed by this integration.  I.e. "osd" or "rhmi"
+	SnitchNamePostFix string `json:"snitchNamePostFix"`
```
jewzaam, line 26:
```suggestion
	SnitchNamePostFix string `json:"snitchNamePostFix,omitempty"`
```

pkg/apis/deadmanssnitch/v1alpha1/deadmanssnitchintegration_types.go
```diff
@@ -0,0 +1,60 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// DeadmansSnitchIntegrationSpec defines the desired state of DeadmansSnitchIntegration
+type DeadmansSnitchIntegrationSpec struct {
+	//reference to the secret containing deadmanssnitch-api-key
+	DmsAPIKeySecretRef corev1.SecretReference `json:"dmsAPIKeySecretRef"`
+
+	//a label selector used to find which clusterdeployment CRs receive a DMS integration based on this configuration
+	ClusterDeploymentSelector metav1.LabelSelector `json:"clusterdeploymentSelector"`
```
jewzaam, line 17:
```suggestion
	ClusterDeploymentSelector metav1.LabelSelector `json:"clusterDeploymentSelector"`
```

pkg/apis/deadmanssnitch/v1alpha1/deadmanssnitchintegration_types.go
```diff
@@ -0,0 +1,60 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// DeadmansSnitchIntegration is the Schema for the deadmanssnitchintegrations API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=deadmanssnitchintegrations,scope=Namespaced
+type DeadmansSnitchIntegration struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   DeadmansSnitchIntegrationSpec   `json:"spec,omitempty"`
```
jewzaam, line 45:
```suggestion
	Spec   DeadmansSnitchIntegrationSpec   `json:"spec"`
```
Is there a use case for an empty spec?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+
+	if dmsi.DeletionTimestamp != nil {
+		for _, clustDeploy := range matchingClusterDeployments.Items {
+			err = r.deleteDMSI(dmsi, &clustDeploy, dmsc)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+		}
+		return reconcile.Result{}, nil
```
jewzaam, line 175:
```suggestion
		logger.Info("Deleting DMS finalizer from dmsi")
		if utils.HasFinalizer(dmsi, DeadMansSnitchFinalizer) {
			utils.DeleteFinalizer(dmsi, DeadMansSnitchFinalizer)
			err = r.client.Update(context.TODO(), dmsi)
			if err != nil {
				logger.Error(err, "Error deleting Finalizer from dmsi")
				return reconcile.Result{}, err
			}
		}

		return reconcile.Result{}, nil
```

Only delete DSMI finalizer after done deleting DMS services for each CD.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+	dmsc := r.dmsclient(dmsAPIKey, localmetrics.Collector)
+
+	matchingClusterDeployments, err := r.getMatchingClusterDeployment(dmsi)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if dmsi.DeletionTimestamp != nil {
+		for _, clustDeploy := range matchingClusterDeployments.Items {
+			err = r.deleteDMSI(dmsi, &clustDeploy, dmsc)
```
jewzaam, line 169:
```suggestion
			err = r.deleteDMSClusterDeploy(dmsi, &clustDeploy, dmsc)
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+		if err != nil {
+			logger.Error(err, "Error deleting Finalizer from dmsi")
+			return err
+		}
+
+		return nil
+	}
+	return nil
+
+}
```
jewzaam, line 480:
```suggestion
```

Duplicate of `deleteDMSClusterDeploy`.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+
+// delete snitches,secrets and syncset associated with the cluster deployment that has been deleted
+func (r *ReconcileDeadmansSnitchIntegration) deleteDMSClusterDeploy(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, clustDeploy *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+
+	var DeadMansSnitchFinalizer string = "dms.managed.openshift.io/deadmanssnitch-" + dmsi.Name
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", clustDeploy.Name, "cluster-deployment.Namespace:", clustDeploy.Namespace)
+	// Delete the dms
+	logger.Info("Deleting the DMS from api.deadmanssnitch.com")
+	snitchName := clustDeploy.Spec.ClusterName + "." + clustDeploy.Spec.BaseDomain + "-" + dmsi.Spec.SnitchNamePostFix
```
jewzaam, line 489: OSD doesn't currently append a post-fix so we need this to be an optional property.  I have a suggestion in the type for this.  If SnitchNamePostFix is not set / is empty don't append `-`.  This change needs to be done for SyncSet and Secret as well.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+	snitches, err := dmsc.FindSnitchesByName(snitchName)
+	if err != nil {
+		return err
+	}
+	for _, s := range snitches {
+		delStatus, err := dmsc.Delete(s.Token)
+		if !delStatus || err != nil {
+			logger.Info("Failed to delete the DMS from api.deadmanssnitch.com")
```
jewzaam, line 497:
```suggestion
			logger.Error("Failed to delete the DMS from api.deadmanssnitch.com")
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+
+		return nil
+	}
+	return nil
+
+}
+
+// delete snitches,secrets and syncset associated with the cluster deployment that has been deleted
+func (r *ReconcileDeadmansSnitchIntegration) deleteDMSClusterDeploy(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, clustDeploy *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
```
jewzaam, line 483: The CRD name this references is `ClusterDeployment`.  I'd prefer we use the full name for the function name and function argument and not shortened.  I won't hold the review for that though.  Also acceptable is the CRD's short name "cd" but that's pretty terse.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+
+	// Delete the referenced secret
+	logger.Info("Deleting DMS referenced secret")
+	err = utils.DeleteRefSecret(clustDeploy.Name+"-"+dmsi.Spec.SnitchNamePostFix+config.RefSecretPostfix, clustDeploy.Namespace, r.client)
+	if err != nil {
+		logger.Error(err, "Error deleting secret")
+		return err
+	}
+
+	logger.Info("Deleting DMS finalizer from clusterdeploy")
```
jewzaam, line 519:
```suggestion
	logger.Info("Deleting DMS finalizer from ClusterDeployment")
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+
+	utils.DeleteFinalizer(clustDeploy, DeadMansSnitchFinalizer)
+	err = r.client.Update(context.TODO(), clustDeploy)
+	if err != nil {
+		logger.Error(err, "Error deleting Finalizer from clustdeploy")
```
jewzaam, line 524:
```suggestion
		logger.Error(err, "Error deleting Finalizer from ClusterDeployment")
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+			return reconcile.Result{}, nil
+
+		}
+
+		if !clustDeploy.Spec.Installed {
+			// Cluster isn't installed yet, return
+			return reconcile.Result{}, nil
+		}
+
+		err = r.dmsAddFinalizer(dmsi, &clustDeploy)
```
jewzaam, line 193: I'm torn about this one.  In general, any update to a watched resource should be followed by exiting the reconcile loop. 
 Every time you update a watched resource a new event is triggered which calls Reconcile. That means;

- dmsAddFinalizer
- createSecret
- createSyncSet

But these are all idempotent and small.  Only issue might be attempted create of the secret or syncset in two threads if the operator supports it.  And that would just mean one fails and is requeued and is a noop for the next reconcile.  Thoughts?  I think this is something I didn't consider with the pagerduty-operator updates.

pkg/utils/utils.go
```diff
@@ -111,7 +114,7 @@ func DeleteSyncSet(name string, namespace string, client client.Client, reqLogge
...
-	reqLogger.Info("Deleting SyncSet", "Namespace", namespace, "Name", name)
+	log.Info("Deleting SyncSet", "Namespace", namespace, "Name", name)
```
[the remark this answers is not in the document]
R-Lawton: Added dmsi and clusterdeployment to log

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+	logger.Info("DMSI finalizer created nothing to do here..: ")
+
+	return nil
+
+}
+
+// create snitch in deadmanssnitch.com with information retrived from dmsi cr as well as the matching cluster deployment
+func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	logger.Info("Checking if snitch already exits Name: ")
```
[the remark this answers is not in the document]
R-Lawton: Cool sounds good to me i added the snitch name on top of whats added with the defualt logger message i.e clusterdeployment etc

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+	}
+	logger.Info("Secret created , nothing to do here...")
+	return nil
+}
+
+//creating the syncset which contain the secret with the snitch url
+func (r *ReconcileDeadmansSnitchIntegration) createSyncset(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd hivev1.ClusterDeployment) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	dmsSecret := cd.Spec.ClusterName + "-" + dmsi.Spec.SnitchNamePostFix + "-" + "dms-secret"
```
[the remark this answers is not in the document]
R-Lawton: The dms configuration doc has this as the layout for the secret to be <cluster-name>-<snitchNamePostfix>-dms-secret

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+	ssName := cd.Spec.ClusterName + "-" + dmsi.Spec.SnitchNamePostFix + "-" + "dms-secret"
```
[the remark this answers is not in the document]
R-Lawton: Sure this can be changed would you think this would cause confusion tho?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,530 @@
...
+
+	matchingClusterDeployments := &hivev1.ClusterDeploymentList{}
+	listOpts := &client.ListOptions{LabelSelector: selector}
+	err = r.client.List(context.TODO(), matchingClusterDeployments, listOpts)
+	return matchingClusterDeployments, err
+}
+
+// Add finalizers to both the deadmanssnitch integreation and the matching cluster deployment
+func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, clustDeploy *hivev1.ClusterDeployment) error {
+	var DeadMansSnitchFinalizer string = "dms.managed.openshift.io/deadmanssnitch-" + dmsi.Name
```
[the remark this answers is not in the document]
R-Lawton: yeah i agree this was taken from the previous way the dms operator was done but happy to have it cohesive

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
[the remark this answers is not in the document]
grdryn: @R-Lawton that's what `config.RefSecretPostfix` is: https://github.com/openshift/deadmanssnitch-operator/blob/master/config/config.go#L21

The reason I suggested it is that `config.RefSecretPostfix` is used in other places for the suffix here.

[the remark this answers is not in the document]
R-Lawton: holy moly i cant read properly apologies i misread your comment before i thought you were referencing something completlety different from the CR yeah sure i with ya now

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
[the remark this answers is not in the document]
grdryn: I think it would be less confusing :see_no_evil: but if you think it would be more confusing, then that's also fair. It's not a big issue either way :slightly_smiling_face:

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
grdryn: @jewzaam is your intention that the OSD DMSI would reconcile the existing snitch & syncset & secret? I thought those were just going to be superseded by the new ones created here (with e.g. an `osd` suffix), and the old ones manually removed?

A problem I see with making the field optional is that if any other DMSI CR also doesn't specify it, then unexpected things will happen (or expected things won't).

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
grdryn: @jewzaam we actually _did_ discuss this a little in the paterduty-operator changes :grin: https://github.com/openshift/pagerduty-operator/pull/90#discussion_r448159838

> Only issue might be attempted create of the secret or syncset in two threads if the operator supports it.

In the above linked conversation, I made the case that there won't be two threads reconciling at the same time (unless we deliberately change things to allow such non-default behaviour), but rather than try to argue the point too much (I might well be wrong), I just changed it (but just that one case that was highlighted, it looks like I didn't consider the other cases...I think that was probably because I was thinking "maybe their point is that a completely separate process, such as hive or another operator, might try to reconcile" since it was on adding the finalizer to the ClusterDeployment; but thinking about it now, I don't think that's what you meant): https://github.com/openshift/pagerduty-operator/pull/90/commits/20f2b769e3bfd128edb90233b0e80abe4518077f

I do see that the existing code has a comment that says "strange things will happen" if the reconcile is not exited early after update (there wasn't any such thing in the pagerduty-operator before the new changes there, and the reconcile wasn't exited after updating the finalizer): https://github.com/openshift/deadmanssnitch-operator/blob/master/pkg/controller/deadmanssnitch/deadmanssnitch_controller.go#L157..L160

So, I'm also torn :slightly_smiling_face: While this operator (before these changes) does return early after adding the finalizer to the ClusterDeployment, it doesn't return early after the creating the secret or syncset. I think to do it properly for the create/update (currently just create happens, I think), we'd need to use `controllerutil.CreateOrUpdate` (which is awesome), and return and requeue if the operation result is not None. I think that would be better done in a separate PR after these changes, WDYT?

@R-Lawton I don't think it would be a huge amount of work to change this particular case to return early after adding finalizers to DMSI and CD resources, WDYT?

pkg/apis/deadmanssnitch/v1alpha1/deadmanssnitchintegration_types.go
R-Lawton: Removed this, Copy and paste error

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
R-Lawton: You only want things to be deleted when the Clusterdeployment is deleted ? Not when the dms integration cr is deleted?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
R-Lawton: I agree i think it would be a good idea. I think a separate PR also make sense as well

jewzaam: Thanks for the reminder of that discussion on the pagerduty-operator!  I was worried I was giving inconsistent feedback.. whew.

That being said, yes I think this is a small thing and yes if we want to change it a separate PR is fine.  The thing that is keeping me torn is that reconcile is against the DMSI, so every OSD will be reconciled on any cluster create, cluster update, cluster delete.  This is such a simple operator and simple resources I don't see it being a big risk but do think we should be mindful of what happens as we scale up.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,517 @@
...
+		if !delStatus || err != nil {
+			logger.Info("Failed to delete the DMS from api.deadmanssnitch.com")
+			return err
+		}
+		logger.Info("Deleted the DMS from api.deadmanssnitch.com")
+	}
+
+	// Delete the SyncSet
+	logger.Info("Deleting DMS SyncSet")
+	err = utils.DeleteSyncSet(clustDeploy.Name+"-"+dmsi.Spec.SnitchNamePostFix+config.RefSecretPostfix, clustDeploy.Namespace, r.client)
```
grdryn, line 442:
```suggestion
	err = utils.DeleteSyncSet(clustDeploy.Name+"-"+dmsi.Spec.SnitchNamePostFix+"-"+config.RefSecretPostfix, clustDeploy.Namespace, r.client)
```

You've removed the hyphen from the beginning of `config.RefSecretPostfix`, so it should be added here.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
grdryn: @R-Lawton I see this is marked as resolved, but doesn't seem to have been changed. Same with the one on line 511.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
grdryn: @R-Lawton I _think_ what @jewzaam is suggesting here should be taken in the context of the previous suggestion on L156, but there might be some info missing between the two:

`r.deleteDMSI` and `r.deleteDMSClusterDeploy` are essentially duplicates of each other, except one deletes the finalizer on the DMSI and the other deletes the finalizer on the ClusterDeployment. When we're cleaning stuff up here and deleting the finalizer from the DMSI, the relevant finalizer should also be removed from the ClusterDeployment. So changing line 156 to call `r.deleteDMSClusterDeploy`, and then adding the removal of the finalizer from the DMSI here (outside the range over corresponding ClusterDeployments, since it only needs to be removed once), means that the `r.deleteDMSI` method can be removed.

Does that sound reasonable/correct? @jewzaam please correct me if you were suggesting something else. :slightly_smiling_face:

pkg/controller/deadmanssnitchintegration/mappers.go
```diff
@@ -0,0 +1,97 @@
...
+func (m ownedByClusterDeploymentToDeadMansSnitchIntegrationsMapper) Map(mo handler.MapObject) []reconcile.Request {
+	relevantClusterDeployments := []*hivev1.ClusterDeployment{}
+	// dmsi := &deadmanssnitchv1alpha1.DeadmansSnitchIntegration{}
+	for _, or := range mo.Meta.GetOwnerReferences() {
+		//log.Info(fmt.Sprintf("TOLOWER%v",or))
+		if or.APIVersion == hivev1.SchemeGroupVersion.String() && strings.ToLower(or.Kind) == "clusterdeployment" {
+			cd := &hivev1.ClusterDeployment{}
+			err := m.Client.Get(context.TODO(), client.ObjectKey{Name: or.Name, Namespace: mo.Meta.GetNamespace()}, cd)
+			if err != nil {
+				continue
```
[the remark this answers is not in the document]
grdryn: @jewzaam on this point, do you have an example to point to that shows what you'd usually do for debug level logging, either in this project or in one of your other projects (assuming there's a consistent pattern that you'd like to follow)?

pkg/utils/utils.go
```diff
@@ -110,8 +117,8 @@ func DeleteSyncSet(name string, namespace string, client client.Client, reqLogge
 	}
 
 	// Only delete the syncset, this is just cleanup of the synced secret.
-	// The ClusterDeployment controller manages deletion of the pagerduty serivce.
-	reqLogger.Info("Deleting SyncSet", "Namespace", namespace, "Name", name)
+	// The ClusterDeployment controller manages deletion of the deadmanssnitch serivce.
+	log.Info("Deleting SyncSet", "Namespace", namespace, "Name", name, "DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", clustDeploy.Name, "cluster-deployment.Namespace:", clustDeploy.Namespace)
```
grdryn, line 121: I think these additional name/namespace values are empty here, since you just created empty structs on L103 and L104. Am I right?

Makefile
```diff
@@ -31,3 +31,7 @@ docker-build:
 .PHONY: docker-push
 docker-push:
 	$(DOCKER_CMD) push ${IMG}
+
+.PHONY: generate-crd
+generate-crd:
```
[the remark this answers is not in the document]
R-Lawton: Ok cool ill change that

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
R-Lawton: Good catch thanks

pkg/utils/utils.go
R-Lawton: Oops sorry im adding more to it now

R-Lawton: This is actually not needed as there are logs for syncsets already in the controller. That tell you the dmsi and cluster deployment

pkg/controller/deadmanssnitchintegration/mappers.go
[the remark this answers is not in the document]
jewzaam: I'd have to search other code, don't have any example.  Thought being simply that if we have an error and ignore it let's log it so we have some trace of what is going on if we're debugging.

pkg/controller/deadmanssnitchintegration/mappers.go
```diff
@@ -0,0 +1,97 @@
...
+type ownedByClusterDeploymentToDeadMansSnitchIntegrationsMapper struct {
+	Client client.Client
+}
+
+func (m ownedByClusterDeploymentToDeadMansSnitchIntegrationsMapper) Map(mo handler.MapObject) []reconcile.Request {
+	relevantClusterDeployments := []*hivev1.ClusterDeployment{}
+	// dmsi := &deadmanssnitchv1alpha1.DeadmansSnitchIntegration{}
+	for _, or := range mo.Meta.GetOwnerReferences() {
+		//log.Info(fmt.Sprintf("TOLOWER%v",or))
+		if or.APIVersion == hivev1.SchemeGroupVersion.String() && strings.ToLower(or.Kind) == "clusterdeployment" {
```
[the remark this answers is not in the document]
jewzaam: Thanks if it's driven by owner refs and we know that will always be singular, +1.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
jewzaam: Yea, without getting into what function does what just need to be sure this does what is expected.
1. if CD is deleted, DMS stuff is cleaned up before finalizer is removed from CD
2. if DMSI is deleted, each CD is cleaned up with finalzier removed from CD before finalizer is removed from DMSI

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
jewzaam: Fair point.  We'll have to be careful about migration then.  If we do not delete older snitches but do create new ones that get synced to the cluster we'll have every old snitch alert that the clusters have gone missing.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
R-Lawton: Perfect sounds good im on the same page

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
```diff
@@ -526,6 +602,7 @@ func verifyOtherSyncSetExists(c client.Client, expected *SyncSetEntry) bool {
 
 	found := false
 	for _, ss := range ssList.Items {
+		fmt.Print("TEST", ss)
```
jewzaam, line 605:
```suggestion
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+	logger.Info("DMSI finalizer created nothing to do here..: ")
+
+	return nil
+
+}
+
+// create snitch in deadmanssnitch.com with information retrived from dmsi cr as well as the matching cluster deployment
+func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	snitchName := cd.Spec.ClusterName + "." + cd.Spec.BaseDomain + "-" + dmsi.Spec.SnitchNamePostFix
```
jewzaam, line 262:
```suggestion
	snitchName := cd.Spec.ClusterName + "." + cd.Spec.BaseDomain
	if dmsi.Spec.SnitchNamePostFix != "" {
		snitchName += + "-" + dmsi.Spec.SnitchNamePostFix
	}
```

Need to support no postfix for snitches, it's an optional argument.  We don't have any postfix for osd today.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+	snitches, err := dmsc.FindSnitchesByName(snitchName)
+	logger.Info("Checking if snitch already exists SnitchName:", snitchName)
```
jewzaam, line 264:
```suggestion
	logger.Info("Checking if snitch already exists SnitchName:", snitchName)
	snitches, err := dmsc.FindSnitchesByName(snitchName)
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+func (r *ReconcileDeadmansSnitchIntegration) createSecret(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, dmsc dmsclient.Client, cd hivev1.ClusterDeployment) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	dmsSecret := cd.Spec.ClusterName + "-" + dmsi.Spec.SnitchNamePostFix + "-" + config.RefSecretPostfix
+	logger.Info("Checking if secret already exits")
+	err := r.client.Get(context.TODO(),
+		types.NamespacedName{Name: dmsSecret, Namespace: cd.Namespace},
+		&corev1.Secret{})
+	if errors.IsNotFound(err) {
+		logger.Info("Secret not found creating secret")
+		snitchName := cd.Spec.ClusterName + "." + cd.Spec.BaseDomain + "-" + dmsi.Spec.SnitchNamePostFix
```
jewzaam, line 314:
```suggestion
		snitchName := cd.Spec.ClusterName + "." + cd.Spec.BaseDomain
		if dmsi.Spec.SnitchNamePostFix != "" {
			snitchName += + "-" + dmsi.Spec.SnitchNamePostFix
		}
```
Since this is done in a few places it'd be best to have a utility function to encapsulate it.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+}
+
+// delete snitches,secrets and syncset associated with the cluster deployment that has been deleted
+func (r *ReconcileDeadmansSnitchIntegration) deleteDMSClusterDeployment(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, clusterDeployment *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+
+	deadMansSnitchFinalizer := "dms.managed.openshift.io/deadmanssnitch-" + dmsi.Name
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", clusterDeployment.Name, "cluster-deployment.Namespace:", clusterDeployment.Namespace)
+	// Delete the dms
+	logger.Info("Deleting the DMS from api.deadmanssnitch.com")
+	snitchName := clusterDeployment.Spec.ClusterName + "." + clusterDeployment.Spec.BaseDomain + "-" + dmsi.Spec.SnitchNamePostFix
```
jewzaam, line 433:
```suggestion
	snitchName := clusterDeployment.Spec.ClusterName + "." + clusterDeployment.Spec.BaseDomain
	if dmsi.Spec.SnitchNamePostFix != "" {
		snitchName += + "-" + dmsi.Spec.SnitchNamePostFix
	}
```

jewzaam:
We need a test for when SnitchNamePostFix is not set as well, forgot to mention.  The snitch name should be just the cluster deployment name +base domain.

pkg/utils/utils.go
```diff
@@ -110,8 +113,7 @@ func DeleteSyncSet(name string, namespace string, client client.Client, reqLogge
 	}
 
 	// Only delete the syncset, this is just cleanup of the synced secret.
-	// The ClusterDeployment controller manages deletion of the pagerduty serivce.
-	reqLogger.Info("Deleting SyncSet", "Namespace", namespace, "Name", name)
+	// The ClusterDeployment controller manages deletion of the deadmanssnitch serivce.
```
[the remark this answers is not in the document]
R-Lawton: it was removed as its a duplicate of the same log its just in the controller instead

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
R-Lawton: I know the postfix was made optional a few commits back but im thinking now it might cause problems down the road with DMS name duplication.

R-Lawton: @jewzaam

R-Lawton: Ok cool ill make the suggested changes

jewzaam:
looks good!
/approve

holding so we can get the initial osd dsmi added (https://github.com/R-Lawton/deadmanssnitch-operator/pull/1) and internal configs done
/hold

deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml
```diff
@@ -0,0 +1,103 @@
...
+                  description: Name is unique within a namespace to reference a secret resource.
+                  type: string
+                namespace:
+                  description: Namespace defines the space within which the secret name must be unique.
+                  type: string
+              type: object
+          required:
+            - clusterDeploymentSelector
+            - dmsAPIKeySecretRef
+            - tags
```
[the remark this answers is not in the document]
jewzaam: From our point of view it's required.  But it could be moved to optional as a backwards compatible change.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+
+	return nil
+
+}
+
+// create snitch in deadmanssnitch.com with information retrived from dmsi cr as well as the matching cluster deployment
+func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	snitchName := cd.Spec.ClusterName + "." + cd.Spec.BaseDomain + "-" + dmsi.Spec.SnitchNamePostFix
+	snitches, err := dmsc.FindSnitchesByName(snitchName)
```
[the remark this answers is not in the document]
jewzaam: +1 this is a point of risk with scaling.  We can assume if the SS and secret exist that the snitch config is good and do nothing.  We have similar concerns in PD operator and I have a WIP branch to change it.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,481 @@
...
+	return nil
+
+}
+
+// create snitch in deadmanssnitch.com with information retrived from dmsi cr as well as the matching cluster deployment
+func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
+	logger := log.WithValues("DeadMansSnitchIntegreation.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	snitchName := utils.DmsSnitchName(cd.Spec.ClusterName, cd.Spec.BaseDomain, dmsi.Spec.SnitchNamePostFix)
+	logger.Info("Checking if snitch already exists SnitchName:", snitchName)
+	snitches, err := dmsc.FindSnitchesByName(snitchName)
```
mwoodson, line 269: This is a call to the DMS api. This is in the loop for every CD in hive.

I believe we need to only check if the snitch is there, if a corresponding syncset doesn't exist for the CD. This will cut down on api calls, and time to run the reconcile loop. I know this opens the operator up to not checking and reconciling what's already there, but I think this risk is worth it.

deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml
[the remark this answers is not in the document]
R-Lawton: Tags are for grouping snitches together making snitches more organized and easier to see. I don't have much of an opinion on whether it should be required. I can change it if you think that best

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -0,0 +1,475 @@
...
+			// Cluster isn't installed yet, return
+			return reconcile.Result{}, nil
+		}
+
+		err = r.dmsAddFinalizer(dmsi, &clusterdeployment)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		err = r.createSnitch(dmsi, &clusterdeployment, dmsc)
```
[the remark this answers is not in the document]
R-Lawton: Only CD's with the correct matching labels will be watched for changes and the same for going through the this

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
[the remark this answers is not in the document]
R-Lawton: Ok sounds good to me

deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml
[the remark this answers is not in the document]
R-Lawton: So will we keep it as required then?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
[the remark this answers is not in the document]
R-Lawton: The new changes to stop making unnecessary API calls to Deadmansnsitch should stop any problems with potential DOSing

deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml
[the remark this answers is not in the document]
jewzaam: In DMS tags are optional, so I guess it's best to be optional in the operator.

jewzaam (approved):

PR #81, opened 2020-10-20
Update yq syntax and fix saas file location

jharrington22:


hack/app_sre_create_image_catalog.sh
```diff
@@ -26,9 +26,11 @@ git clone \
 # remove any versions more recent than deployed hash
 REMOVED_VERSIONS=""
 if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
+    # Can't use the var _OPERATOR_NAME here in the saas file path because app-interface has a typo with deadmanssntich being deadmansnitch.
+    # Until that is fixed hard coding
     DEPLOYED_HASH=$(
-        curl -s "https://gitlab.cee.redhat.com/service/saas-osd-operators/raw/master/${_OPERATOR_NAME}-services/${_OPERATOR_NAME}.yaml" | \
-            docker run --rm -i quay.io/app-sre/yq -r '.services[]|select(.name="${_OPERATOR_NAME}").hash'
+        curl -s 'https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-deadmansnitch-operator.yaml' | \
```
jewzaam, line 32: For posterity, the name of the file is wrong, doesn't match the name of the operator (missing one `s`) so using `${_OPERATOR_NAME}` isn't possible.

2uasimojo: We're going to need to fix that: boilerplate will rely on the names matching. I'm guessing renaming the file is going to be, like, five steps. @maorfr?

jharrington22: I'm on it, just wanted to get it deployable in production again, if we were to promote now it would break. 

Fix for app-interface: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10829

2uasimojo: Yup, cool. Comments there.

maorfr: @2uasimojo did it end up being a single step? :wink: 
and just for visibility on this PR as well: https://github.com/openshift/aws-account-operator/pull/493

2uasimojo: @maorfr yes, people were shepherding the process through, so there was just the one app-interface MR to rename and then a fup here to switch over.

PR #83, opened 2020-10-20
Use var for operator name in hack/app_sre_create_image_catalog.sh

jewzaam:


jewzaam:
https://github.com/openshift/deadmanssnitch-operator/pull/82 wins :D

PR #88, opened 2020-11-03
Move CRD fixup hack for 3.11

2uasimojo:
Due to an OLM bug in OCP 3.11, the generated CRD needs some tweaking. Previously this was being done via a bespoke script invoked through a go:generate directive. After boilerplate adoption, this script was doing `operator-sdk generate` redundantly (it was already part of the `op-generate` target).

This commit removes the script and the redundant `operator-sdk generate`s, and adds a temporary `make` target to do the CRD fixup. All of this will go away once we no longer need to run on 3.11.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,3 +4,14 @@ include boilerplate/generated-includes.mk
 boilerplate-update:
 	@boilerplate/update
 
+# ==> HACK ==>
+# We have to tweak the YAML to run this on 3.11.
+# We'll add a target to do that, then override the `generate` target to
+# call it.
+.PHONY: crd-fixup
+crd-fixup:
+	yq d -i deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml spec.validation.openAPIV3Schema.type
+
+# NOTE: go-generate is a no-op at the moment.
+generate: op-generate crd-fixup go-generate
+# <== HACK <==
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -4,3 +4,14 @@ include boilerplate/generated-includes.mk
 boilerplate-update:
 	@boilerplate/update
 
+# ==> HACK ==>
+# We have to tweak the YAML to run this on 3.11.
+# We'll add a target to do that, then override the `generate` target to
+# call it.
+.PHONY: crd-fixup
+crd-fixup:
+	yq d -i deploy/crds/deadmanssnitch.managed.openshift.io_deadmanssnitchintegrations_crd.yaml spec.validation.openAPIV3Schema.type
```
cblecker, line 13: Do we have a standardized version of yq?

georgettica:
Is this change going to be in the boilerplate? I am asking as another operator has concerns for the CRD deployment

Makefile
2uasimojo: It's in the [boilerplate image](https://github.com/openshift/boilerplate/blob/master/config/build.sh#L45-L50)

2uasimojo:
> Is this change going to be in the boilerplate? I am asking as another operator has concerns for the CRD deployment

Yeah, that's a good question. I'd like to solve this one by hand the first time around to make sure it's working, and then we can look at making it generic (we'll have to discover which files to apply the fixup to).

It's temporary (only as long as 3.11 lives) so not worth putting a whole lot of effort into.

PR #92, opened 2020-12-01
OSD-5589 Remove DMS from non-matching clusters

mrbarge:
### What type of PR is this?

Bug/Feature

### What this PR does / why we need it?

[OSD-5589](https://issues.redhat.com/browse/OSD-5589) requests that the DMS setup should be removed for any cluster which does not match the [DMSI clusterDeploymentSelector](https://github.com/openshift/deadmanssnitch-operator/blob/f9cfccd45c4f99b9768c2df8d44c96f2457b169b/hack/olm-registry/olm-artifacts-template.yaml#L56). This PR revises the reconcile logic of the DMSI controller such that any cluster which does not match the DMSI cluster selector will have its DMS setup removed.

Presently, prior to this PR, the behaviour of the DMS reconciler is:

- For all `clusterdeployments` (matching & non-matching): 
  - if any `clusterdeployment` is being deleted, remove DMS.
- For all matching `clusterdeployments`: 
  - If the cluster isn't yet installed and isn't being deleted, setup DMS.

In this PR, the behaviour is changed to:

- For all `clusterdeployments`:
  - If the `clusterdeployment` DOES NOT match the selector and has DMS setup, remove DMS.
  - If the `clusterdeployment` DOES match the selector and is being deleted, remove DMS.
  - If the `clusterdeployment` DOES match the selector and isn't yet installed, do nothing.
  - If the `clusterdeployment` DOES match the selector and is neither being deleted nor pending install, setup DMS.

Tests have been added to test these situations.

### Which Jira/Github issue(s) this PR fixes?

[OSD-5589](https://issues.redhat.com/browse/OSD-5589)

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -180,24 +180,30 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 	}
 
 	for _, clusterdeployment := range allClusterDeployments.Items {
-		if clusterdeployment.DeletionTimestamp != nil {
+
+		// Check if the cluster matches the requirements for needing DMS setup
+		clusterMatched := false
+		for _, matchingClusterDeployment := range matchingClusterDeployments.Items {
+			if clusterdeployment.UID == matchingClusterDeployment.UID {
+				clusterMatched = true
+				break
+			}
+		}
+
+		if !clusterMatched || clusterdeployment.DeletionTimestamp != nil{
+			// The cluster does not match the criteria for needing DMS setup
 			if utils.HasFinalizer(&clusterdeployment, deadMansSnitchFinalizer) {
+				// The cluster has an existing DMS setup, so remove it
 				err = r.deleteDMSClusterDeployment(dmsi, &clusterdeployment, dmsc)
 				if err != nil {
 					return reconcile.Result{}, err
 				}
 			}
-		}
-	}
-
-	for _, clusterdeployment := range matchingClusterDeployments.Items {
-		if clusterdeployment.DeletionTimestamp != nil {
-			// no action required, as this should be handled by the all function above
 			continue
 		}
 
 		if !clusterdeployment.Spec.Installed {
-			// Cluster isn't installed yet, continue
+			// The cluster isn't installed yet, so don't setup DMS yet either
 			continue
 		}
 
@@ -215,6 +221,7 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+
 		err = r.createSyncset(dmsi, clusterdeployment)
 		if err != nil {
 			return reconcile.Result{}, err
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
@@ -40,6 +40,7 @@ const (
 	testTag                           = "test"
 	testAPIKey                        = "abc123"
 	testOtherSyncSetPostfix           = "-something-else"
+	testUID							  = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
 	snitchNamePostFix                 = "test-postfix"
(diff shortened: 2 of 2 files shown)


config/config.go
```diff
@@ -24,4 +24,10 @@ const (
 	// ClusterDeploymentManagedLabel is the label the clusterdeployment will have that determines
 	// if the cluster is OSD (managed) or not
 	ClusterDeploymentManagedLabel string = "api.openshift.com/managed"
+	// NoAlertsLabel is the label the clusterdeployment will have if the cluster is not meant
+	// to be generating alerts (including DMS)
+	NoAlertsLabel string = "ext-managed.openshift.io/noalerts"
+	// LegacyNoAlertsLabel is the legacy label the clusterdeployment will have if the cluster is
+	// not meant to be generating alerts (including DMS)
+	LegacyNoAlertsLabel string = "api.openshift.com/noalerts"
```
jewzaam, line 32: These are used only by tests and there is on logic that actually turns off DMS when they are present.  This logic was removed and moved to DSMI config for our envs (ironically this is why we noticed this issue needed to be fixed).

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -180,24 +180,43 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
...
-			// no action required, as this should be handled by the all function above
+			// The clusterdeployment is being deleted, so remove DMS
+			if utils.HasFinalizer(&clusterdeployment, deadMansSnitchFinalizer) {
+				err = r.deleteDMSClusterDeployment(dmsi, &clusterdeployment, dmsc)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+			}
 			continue
 		}
```
jewzaam, line 216: Could collapse these into a single block, 
```
if !clusterMatched || clusterdeployment.DeletionTimestamp != nil {
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
```diff
@@ -307,6 +323,50 @@ func TestReconcileClusterDeployment(t *testing.T) {
...
+				testSecret(),
+				noalertsManagedClusterDeployment(),
+				testDeadMansSnitchIntegration(),
+			},
+			expectedSyncSets: &SyncSetEntry{},
+			expectedSecret:   &SecretEntry{},
+			verifySyncSets:   verifyNoSyncSet,
+			verifySecret:     verifyNoSecret,
+			setupDMSMock: func(r *mockdms.MockClientMockRecorder) {
+				r.Create(gomock.Any()).Times(0)
```
jewzaam, line 338: I am confused by this test, I would expect it to try to setup DMS since the "noalerts" labels have no meaning anymore in the code and the test DMSI only matches on managed=true.

jewzaam (changes requested):

config/config.go
mrbarge: Agree on this point - I've removed this now. (see my comment on the topic of the confusing test)

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
mrbarge: Thanks, that does indeed make things simpler - I've done this now.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
mrbarge: Yeah, you're totally correct here. This test was only working because it was using a non-managed clusterdeployment with no DMS finalizer, and alerts didn't factor into it at all. Thanks for that catch.

In the branch now, I have created a new test as a replacement - one that will return a non-managed cluster with a DMS finalizer in it. This will exercise the new behaviour that we want this controller to take - for a cluster that doesn't match the `clusterDeploymentSelector` but has a DMS finalizer  -> delete the DMS setup. I think this is what we actually want here.

Thanks a tonne for noticing this - very much a :facepalm: moment!

jewzaam: Fwiw I spent time in the PD operator to refactor tests and was considering doing it for DMS.  Not asking for that change here but may be something to consider in the future.

jewzaam (approved):

PR #94, opened 2020-12-03
remove DMS from cluster when addon requiring it is uninstalled

philbrookes:
Check for the removal of the label on clusters that still have the finalizer, and remove DMS from any clusters that match this criteria.

In the scenario where a cluster had an addon which required DMS and later the addon is removed, but the cluster is not deleted - the DMS operator will not remove the syncset for that cluster, meaning that the DMS will still be firing for this cluster.

This PR will find and clean up clusters in that scenario.


philbrookes:
https://github.com/openshift/deadmanssnitch-operator/pull/92/files

PR #100, opened 2021-02-23
[OSD-6376] update dms operator to work with cluster relocation

bmeng:


pkg/utils/utils.go
```diff
@@ -39,6 +40,29 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// IsRelocating indicates that the cluster deployment is being relocating
+func IsRelocating(cd *hivev1.ClusterDeployment) bool {
+	hasAnnotation := HasAnnotation(cd, hiveconst.RelocateAnnotation, strings.HasSuffix, string(hivev1.RelocateOutgoing), string(hivev1.RelocateIncoming))
```
mrbarge, line 61: Maybe also include `hive1.RelocateComplete` as part of this check - my understanding is that the clusterdeployment will receive that annotation right before deletion on the outgoing Hive.

bmeng: yeah, I am a bit confuse about the Complete status.
During my testing, the annotation will be added/removed rapidly. I cannot figure out each phase of it...

mrbarge: So the `/complete` annotation will be set on the outgoing Hive's clusterdeployment when the cluster relocation has finished, but prior to the clusterdeployment being deleted. It should remain in a `complete` state up until the point that the clusterdeployment is deleted.

The easiest way to replicate this in a test is to add a fake finalizer to the `clusterdeployment` prior to relocating it. That way, it's not going to get deleted straight away, and when you look at the `clusterdeployment` you'll see that it will have  that annotation present.

Because there will be multiple operators each individually handling the way that they remove their finalizer from the `clusterdeployment` during a relocation event, it is going to be quite possible that the `clusterdeployment` could be in that `complete` annotation state for quite a while (ie. imagine if the certman-operator took 5 minutes to do its own cleanup). So yeah, I'd say it's safest to include it in the `IsRelocating` test in order to prevent the chance/race condition that the outgoing DMS operator deletes the DMS service after the incoming DMS operator has reconciled its existence.

PR #101, opened 2021-03-09
ignore fake clusters

bmeng:
fixes https://issues.redhat.com/browse/OSD-6346


mrbarge:
Could the `DeadMansSnitchIntegration` be an alternate way of doing this?

ie. adding a new entry to the `matchExpressions` for our existing DMSI's on Hive:
```
  clusterDeploymentSelector:
    matchExpressions:
    - key: hive.openshift.io/fake-cluster
      operator: NotIn
      values:
      - "true"
```

bmeng:
Have tried with the DMSI, looks like it will only watch the `Labels` in ClusterDeployment, but the fake cluster indicator appears in the `Annotations`.

jewzaam:
Sounds reasonable, will approve.  Putting on hold for the question of label so this is configuration and not code to implement.  If we decide to go with the annotation.

/approve
/hold

bmeng:
close this one, use #102 instead

PR #102, opened 2021-03-15
use the DMSI to configure ignore fake clusters

bmeng:
an alternative way for https://github.com/openshift/deadmanssnitch-operator/pull/101


jewzaam:
Note, this is for https://issues.redhat.com/browse/OSD-6346

jewzaam (changes requested):
Recommend adding tests with ClusterDeploymentAnnotationsToSkip not set on DMSI and set to an empty set.  Looks like there's the test only with fake-cluster annotation set to skip for all tests.

jewzaam:
Approving and putting on hold for further review, in case I can't personally review with addition of tests or team decides we're OK with tests as-is.

/approve
/hold

jewzaam:
Thanks for updating tests.

/lgtm

mrbarge:
I've had a read through the change and it looks good to me. I don't think an /lgtm will work as I'm not in `OWNERS` but nevertheless, PR looks fine. :+1:

jewzaam:
@mrbarge anybody can /lgtm.  only /approve needs to be in OWNERS.  I take your comment as a lgtm and am adding the tag.

/lgtm

bmeng:
yeah, but there is a `hold` which blocks the merge I think...

PR #108, opened 2021-04-13
Boilerplate update with common appsre pipeline

2uasimojo:
- Automated boilerplate update
- Fixups to accommodate ^
- Another boilerplate update to bring in fixes discovered late
- And another of the same

See individual commits for details.

[OSD-6851](https://issues.redhat.com/browse/OSD-6851)


2uasimojo:
/hold

py3 doesn't exist on the jenkins server, so the appsre pipeline will fail as written. We're working on getting a py3 container image we can use to invoke the bundle generator.

2uasimojo:
This is ready for review. Leaving hold so we can merge these in a controlled order.

PR #117, opened 2021-08-05
Make operator hibernation aware

sedroche:
Jira: https://issues.redhat.com/browse/OSD-6924 - Make DMSO hibernation aware
Docs: https://github.com/openshift/hive/blob/master/docs/hibernating-clusters.md

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -212,22 +212,39 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 			return reconcile.Result{}, err
 		}
 
-		err = r.createSnitch(dmsi, &clusterdeployment, dmsc)
+		secretExist, syncSetExist, err := r.snitchResourcesExist(dmsi, &clusterdeployment)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		clusterIsHibernating := clusterdeployment.Spec.PowerState == hivev1.HibernatingClusterPowerState
 
-		err = r.createSecret(dmsi, dmsc, clusterdeployment)
-		if err != nil {
-			return reconcile.Result{}, err
+		if clusterIsHibernating {
+			if secretExist || syncSetExist {
+				err := r.deleteDMSClusterDeployment(dmsi, &clusterdeployment, dmsc)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+		} else {
+			if !secretExist || !syncSetExist {
+				err = r.createSnitch(dmsi, &clusterdeployment, dmsc)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+
+				err = r.createSecret(dmsi, dmsc, clusterdeployment)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+
+				err = r.createSyncset(dmsi, clusterdeployment)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+			}
 		}
-
-		err = r.createSyncset(dmsi, clusterdeployment)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-
 	}
+
 	log.Info("Reconcile of deadmanssnitch integration complete")
 
 	return reconcile.Result{}, nil
@@ -365,6 +382,32 @@ runbook: https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_
 
 	logger.Info("Snitch created nothing to do here.... ")
 	return nil
+}
+
+// snitchResourcesExist checks if the associated cluster resources for a snitch exist
+func (r *ReconcileDeadmansSnitchIntegration) snitchResourcesExist(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment) (bool, bool, error) {
+	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+
+	dmsSecret := utils.SecretName(cd.Spec.ClusterName, dmsi.Spec.SnitchNamePostFix)
+	logger.Info("Checking if secret exists")
+	secretExist := false
+	err := r.client.Get(context.TODO(),
+		types.NamespacedName{Name: dmsSecret, Namespace: cd.Namespace},
+		&corev1.Secret{})
+	if err != nil && !errors.IsNotFound(err) {
+		return false, false, err
+	}
+	secretExist = !errors.IsNotFound(err)
+
+	logger.Info("Checking if syncset exists")
+	syncSetExist := false
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: dmsSecret, Namespace: cd.Namespace}, &hivev1.SyncSet{})
+	if err != nil && !errors.IsNotFound(err) {
+		return secretExist, false, err
+	}
+	syncSetExist = !errors.IsNotFound(err)
+
+	return secretExist, syncSetExist, nil
 }
 
 //Create secret containing the snitch url
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
@@ -95,6 +95,28 @@ func testSecret() *corev1.Secret {
 	return s
 }
 
+// return a syncset that matches the secret found in the hive namespace
+func testSyncSet() *hivev1.SyncSet {
+	ss := &hivev1.SyncSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testClusterName + "-" + snitchNamePostFix + "-" + config.RefSecretPostfix,
+			Namespace: testNamespace,
+		},
+	}
+	return ss
+}
+
+// return a secret that matches the secret in syncset secret ref
+func testSecretRef() *corev1.Secret {
+	s := &corev1.Secret{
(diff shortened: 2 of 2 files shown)


jewzaam (changes requested):
This logic should check first if there is a secret and/or SS before checking power state.  If secret and/or SS exist and power state = hibernating then delete dms, secret, and ss.  Merging it into each of what was "create" functions before results in massive number of calls out to DMS service.  We must reduce this to only when work might need to be done.  Existance of secret and SS are an indicator of state that either something needs cleaned up or that a DMS service exists.

So logic is:
if (installed AND hibernating AND (exist secret OR exist SS)) then call DMS to check for a snitch, delete snitch if found, delete secret, delete SS.

sedroche:
> This logic should check first if there is a secret and/or SS before checking power state. If secret and/or SS exist and power state = hibernating then delete dms, secret, and ss. Merging it into each of what was "create" functions before results in massive number of calls out to DMS service. We must reduce this to only when work might need to be done. Existance of secret and SS are an indicator of state that either something needs cleaned up or that a DMS service exists.
> 
> So logic is:
> if (installed AND hibernating AND (exist secret OR exist SS)) then call DMS to check for a snitch, delete snitch if found, delete secret, delete SS.

Thanks @jewzaam. I have made the suggested changes, can you take another look please

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -212,6 +212,26 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 			return reconcile.Result{}, err
 		}
 
+		if clusterdeployment.Spec.PowerState == hivev1.HibernatingClusterPowerState {
```
jewzaam, line 215: This will run every reconcile loop.  At a minimum we get 2x lookup of SS, 2x lookup of secret, and a log statement every time it runs.

Recommend pull the logic in deleteSnitch for checking if a delete needs to be done out and before this block.  It could be used to control if create or delete functions are called, optimize everything a bit.

jewzaam (changes requested):
There are no unit tests added to check this functionality works as expected.

sedroche:
@jewzaam I have made the changes requested in https://github.com/openshift/deadmanssnitch-operator/pull/117#discussion_r685197820 - https://github.com/openshift/deadmanssnitch-operator/pull/117/commits/f1db50e4be91960d8f934aaf4c773e2881ea0e14
Also:
- Didn't realise there was utily funcs for deleting secret and syncset so changed to using them - https://github.com/openshift/deadmanssnitch-operator/pull/117/commits/4d73c8867195d7974285ea972b2c8583613bd15e
- Test for hibernation - https://github.com/openshift/deadmanssnitch-operator/pull/117/commits/ba3311f76000c21c47c0ae5080f362e5e9ce1c47

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -367,6 +395,64 @@ runbook: https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_
...
+		&corev1.Secret{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	logger.Info("Checking if syncset exists")
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: dmsSecret, Namespace: cd.Namespace}, &hivev1.SyncSet{})
```
jewzaam, line 415: This will not be called if the secret doesn't exist and the result returned will be `false` indicating there is no DMS snitch to cleanup.  If secret OR syncset exist need to treat as though the snitch _does_ exist.  Else we have an inconsistent state for deletion.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -212,22 +212,50 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
...
+				dmsSecret := utils.SecretName(clusterdeployment.Spec.ClusterName, dmsi.Spec.SnitchNamePostFix)
+				err = utils.DeleteRefSecret(dmsSecret, clusterdeployment.Namespace, r.client)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+
+				err = utils.DeleteSyncSet(dmsSecret, clusterdeployment.Namespace, r.client)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
```
jewzaam, line 237: Reuse existing delete logic.  `deleteDMSClusterDeployment(dmsi, clusterdeployment, dmsc)`

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
```diff
@@ -342,6 +372,39 @@ func TestReconcileClusterDeployment(t *testing.T) {
...
+				snitchURL:                testSnitchURL,
+				clusterDeploymentRefName: testClusterName,
+			},
+			verifySyncSets: verifyNoSyncSet,
+			verifySecret:   verifyNoSecret,
+			setupDMSMock: func(r *mockdms.MockClientMockRecorder) {
+				r.FindSnitchesByName(gomock.Any()).Return([]dmsclient.Snitch{
+					{Token: testSnitchToken},
+				}, nil)
+				r.Delete(gomock.Any()).Return(true, nil)
```
jewzaam, line 400:
```suggestion
				r.FindSnitchesByName(gomock.Any()).Return([]dmsclient.Snitch{
					{Token: testSnitchToken},
				}, nil).Times(1)
				r.Delete(gomock.Any()).Return(true, nil).Times(2)
```

I might have the counts wrong, but you need to verify the mocks are actually called an expected number of times.  We expect to find the snitch once before delete.  And expect to issue a delete on the syncset and secret.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
```diff
@@ -342,6 +372,39 @@ func TestReconcileClusterDeployment(t *testing.T) {
...
+
+				r.Create(gomock.Any()).Times(0)
+				r.Update(gomock.Any()).Times(0)
+				r.CheckIn(gomock.Any()).Times(0)
+				r.Delete(gomock.Any()).Times(0)
```
jewzaam, line 405:
```suggestion
```
We do expect deletes, this is not needed?

jewzaam (changes requested):

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
sedroche: :disappointed: 
Thanks @jewzaam 
Changes are in https://github.com/openshift/deadmanssnitch-operator/pull/117/commits/cf9f7723f8b7cdd77fd694518fbd32ebfebb0bee
I have changed the signature a bit to handle delete if `secretExist || syncSetExist` and create if `!secretExist || !syncSetExist`
Not sure if there is nicer way to define write the logicg

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
sedroche: I had seen that function, and it gets called if ClusterDeployment or DMSIntegration are deleted, updating the finalizers in the process.
I didn't think this situation should do that as we still want a snitch once we switch back out of hibernation. wdyt?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
sedroche: `Times(1)` is implied, so there is no need to check in that case. I have added the checks back in for consistency with the rest of the tests though - https://github.com/openshift/deadmanssnitch-operator/pull/117/commits/fddfa6c29c5263260d2d365bcccdead8cca7176e (This is just the dms client mock though so we only fetch once and delete once.)


The secret and syncset deletion is verified with https://github.com/openshift/deadmanssnitch-operator/blob/fddfa6c29c5263260d2d365bcccdead8cca7176e/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go#L394-L395

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller_test.go
sedroche: This checks whether any extra calls to delete occur outside the expected. I have done it this was to keep consistency with the other tests but might be worth switching to using `gomock.InOrder` to verify this.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
jewzaam: If the cluster is hibernated it's not setup with DMS.  Which makes me think we could approach this a different way.  If we had a label on the CD we could update the DMSI selector to exclude clusters that are hibernated.  Then no special logic in the operator.  I don't think there's a valid reason to keep the finalizer around when there isn't any DMS service.  If someone issued a delete on the CD when hibernated, DMS shouldn't be involved.

sedroche: @jewzaam Ok, cheers. As for this work, I have made that final change to use `deleteDMSClusterDeployment(dmsi, clusterdeployment, dmsc)` and tidied up the commit history.

As for the selector, it's a fair suggestion. I did check a hibernating ClusterDeployment CR in stage and there is currently no label applied - so that would be work needed in OCM side.
Pros for this approach is if we need to refine the hibernation in the future, the code is here and just needs updating. This would be a decision for the dms operator team to make. wdyt?

sedroche: @jewzaam (reping on above :smile: )

PR #118, opened 2021-08-05
OSD-7196 Disable DMS for "fake" OSD clusters

ninabauer:
Adds additional "managed.openshift.com/fake" annotation to skip.

Fixes logic when there are multiple annotations to skip.


NautiluX (commented):
/lgtm

PR #119, opened 2021-08-11
Updated MD5 hashing to SHA224

robotmaxtron:
Updating a weak cypher identified by OSD Operator Threat Model Analysis.

### DIFF
--- a/pkg/utils/secrets.go
+++ b/pkg/utils/secrets.go
@@ -2,7 +2,7 @@ package utils
 
 import (
 	"context"
-	"crypto/md5"
+	"crypto/sha256"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


jewzaam (changes requested):
No description on why we're doing this change.
And where is pkg/utils/secrets even used?  From my read of things it can simply be removed along with tests.

robotmaxtron:
@jewzaam I should've added it earlier and has now been updated. If you'd rather see the tests removed entirely, I can do that if that's preferred.

jewzaam:
> @jewzaam I should've added it earlier and has now been updated. If you'd rather see the tests removed entirely, I can do that if that's preferred.

Thanks for the description update.  

But, I don't see where pkg/util/secrets is used so my question is can it simply be deleted and remove md5 dependency from the repo?

PR #121, opened 2021-10-29
tear down snitch if customer has stopped their instances

yithian:
addresses [OSD-8505](https://issues.redhat.com/browse/OSD-8505)

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/openshift/deadmanssnitch-operator/config"
 	deadmanssnitchv1alpha1 "github.com/openshift/deadmanssnitch-operator/pkg/apis/deadmanssnitch/v1alpha1"
+	dmsAws "github.com/openshift/deadmanssnitch-operator/pkg/clients/aws"
 	"github.com/openshift/deadmanssnitch-operator/pkg/dmsclient"
 	"github.com/openshift/deadmanssnitch-operator/pkg/localmetrics"
 	"github.com/openshift/deadmanssnitch-operator/pkg/utils"
@@ -40,6 +41,7 @@ const (
 	deadMansSnitchAPISecretKey    = "deadmanssnitch-api-key"
 	DeadMansSnitchFinalizerPrefix = "dms.managed.openshift.io/deadmanssnitch-"
 	SREUsernamePrefix             = "SRE-"
+	deadmanssnitchAwsSecretName   = "deadmanssnitch-operator-aws-credentials"
 )
 
 // Add creates a new DeadmansSnitchIntegration Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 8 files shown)


yithian:
cc @csheremeta

yithian:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/27848

this MR will create a secret in vault for the needed AWS credentials and provide them to the hive clusters

csheremeta:
@jharrington22 yes, it's ready!

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -226,7 +235,42 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 				}
 			}
 		} else {
-			if instancesAreRunning(clusterdeployment) {
+			if clusterIsNotHibernating(clusterdeployment) {
+				// try to stop CHGM noise from AWS
```
mrbarge, line 239: I have some concerns with this, because if my understanding of the DMS operator is correct (disclaimer, it's been a while) this change would cause the DMS operator to check instance status in AWS on every reconcile, for every cluster that has no other reason to not be in DMS.

Given a Hive shard could have upwards of several hundred clusters on it, that's a heavy operation to perform every reconcile. Especially considering that causes for DMS operator reconcile can include:

- any single clusterdeployment CR changing (eg. changing a label)
- a select portion of any cluster's selective syncsets or secrets changing for any reason (only the syncsets owned by a CD, so the DMS & PD & Control Plane certs syncsets)

Do we have a rough idea of how long it might take for it to do a full pass over several hundred clusters? And could it successfully do that in under the time it takes for DMS to trigger a CHGM after instances are stopped? (15-29 mins, depending)

If I am instead just reading this wrong and it would not do this for every cluster, my apologies!

yithian: You're correct, CHGM will be checked every reconcile with this patch. It would be an increase in load for the hive clusters / DMS-o but I'm not sure how to do this check without, well, checking.

The only thing I can think of is to build a service that provides a public-facing (or at least deadmanssnitch-facing) webhook that would stand in between deadmanssnitch and PagerDuty to:
* receive an alert from deadmanssnitch
* do the CHGM checks
* drop the alert if the customer stopped their instances
* create a pagerduty incident otherwise

I think this webhook model would require us to create a secure, public-internet-facing service (with information from the clusterdeployments and access to AWS credentials that can jump into _any_ AWS OSD account) and somewhere to run it with high availability and redundant monitoring (since if it goes down, DMS by definition wouldn't be able to alert us). Granted, those are probably all solvable problems, but would take more than the two engineer-weeks that we had between @csheremeta and I.

So all that to say that I'd rather spend CPU cycles on this problem than have humans continue manually responding to CHGM alerts that are almost pure noise :)

We tested this on hive-stage-01 with a test cluster and for what it's worth, the snitch got torn down practically the same time the instances started shutting down.

fahlmant: I think the service you describe is a much better option, and I outlined a similar one in the design doc that resulted in this PR.

csheremeta: For what it's worth, checking the instance status on every cluster is *not* code that Alex and I introduced. The only thing we did here was change the name of the function that was already checking the instance status for every cluster (because the old name, `instancesAreRunning`, is not exactly what that function was checking). This function rename should not be a blocking change for this PR.

And while the webhook service Alex described is a better option, like Alex said, we had 2 engineers with a week to spare. I would like to see this code merged so that humans no longer have to respond to these alerts when they are the cause of customers stopping their instances, and we can work on improving the way this is built as part of a larger effort.

yithian: From a purely technical standpoint, I would agree. But it also requires a significantly higher level of effort and coordination than what could reasonably be accomplished by two engineers in a week.

Unfortunately, I think time very much is a factor here. As I described in the proposal doc, we had 298 CHGM incidents in August (for ease of math, let's call it 300). That's ~10 per day and ~2 per shift. If we're planning to scale 10x in a year, that means we're expecting 20 CHGM incidents per shift by then and (assuming linear growth) 10 per shift 6 months from now.

If we'd rather implement a CHGM webhook service instead, honestly I'd fully support that. Do we know which functional team(s) would be tasked with creating it? How soon can they get started and what would be taken off their plate to do so?

csheremeta: ignore me, I was confused :)

jharrington22:
@mrbarge @yithian @csheremeta @fahlmant lets move to the card for some of this discussion which isn't relevant to DMS like a service, might be confusing to folks who don't have context. [OSD-8505]( https://issues.redhat.com/browse/OSD-8505) thanks!

PR #122, opened 2021-12-07
OSD-9238 Added var for Fedramp

luis-falcon:
This PR adds a deployment parameter to the OLM template so that AppSRE can deploy the operator in different govcloud vs. non-govcloud environments.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -16,6 +16,8 @@ parameters:
   value: '["None"]'
 - name: DEADMANSSNITCH_OSD_TAGS
   required: true
+- name: FEDRAMP
+  value: "false"
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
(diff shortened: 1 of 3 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -16,6 +16,8 @@ parameters:
   value: '["None"]'
 - name: DEADMANSSNITCH_OSD_TAGS
   required: true
+- name: FEDRAMP
+  value: "${FEDRAMP}"
```
dustman9000, line 20: This needs to default to "false".
```suggestion
  value: "false"
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -123,6 +126,13 @@ type ReconcileDeadmansSnitchIntegration struct {
 func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling DeadmansSnitchIntegration")
+
+	if len(os.Getenv("FEDRAMP")) > 0 {
```
dustman9000, line 130: Should check for zero length here:
```suggestion
	if len(os.Getenv("FEDRAMP")) == 0 {
```

PR #124, opened 2021-12-20
feat(hibernation): add more conditions [OSD-9349]

georgettica:


georgettica:
Not only do the merge conflicts exist, but the change @jharrington22 noted does.the same
/Close

PR #125, opened 2021-12-20
OSD-9349:  bumped hive api scheme and support for ResumingOrRunning hibernation

ramonbutter:
This will change the hibernation condition check of a cluster that requires a DMS integration. 
With the latest hive releases, the hibernation states `running` and `resuming` were merged into a new state `ResumingOrRunning`.
The PR makes DMSO aware of that


yashvardhan-kukreja:
@ramonbutter thanks for this! this looks good to me!

Do you mind also introducing a [INFO] level log [here](https://github.com/openshift/deadmanssnitch-operator/blob/master/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go#L404) stdout-ing "checking snitchResources" (or something like that).

yashvardhan-kukreja:
Also, there's another concern,
[In this section of code](https://github.com/openshift/deadmanssnitch-operator/blob/f444702277c0a40122206662194922ae59964746/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go#L435-L459), the case `if err != nil && !errors.IsNotFound(err)` is not handled the way it should ve been.

Do you mind taking a look at that? :)

fahlmant:
Please add a description

yashvardhan-kukreja (approved):
thanks @ramonbutter, lgtm!

PR #126, opened 2021-12-22
OSD-9349: fix print of boolean

ramonbutter:


pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -130,7 +130,7 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 	if len(os.Getenv("FEDRAMP")) == 0 {
 		reqLogger.Info("FEDRAMP environment variable unset, defaulting to false")
 	} else {
-		reqLogger.Info("running in FedRAMP environment: %b", fedramp)
+		reqLogger.Info(fmt.Sprintf("running in FedRAMP environment: %t", fedramp))
```
georgettica, line 133:
```suggestion
		reqLogger.Infof("running in FedRAMP environment: %t", fedramp)
```

ramonbutter: doesn't exist for that logger :( 
used just the same as here: https://github.com/openshift/deadmanssnitch-operator/blob/f2248a13941f24afea7c5c6c58f127777bbf3466/cmd/manager/main.go#L42

georgettica: your'e right
it's the logger I like :)

georgettica:
nvm

PR #129, opened 2022-01-07
Attempt to verify PowerState before checking legacy conditions

jharrington22:
We can now rely on the ClusterDeployment Status field `PowerState` to determine if a cluster is running or not, this field was introduced in https://github.com/openshift/hive/pull/1604

We only check legacy conditions if that field doesn't exist which may be because the running version of Hive does not yet contain support for that field. 

We are removing checking legacy Hibernation condition reason `Resuming` because during that state the cluster is not "Running" and may not check in with Deadmanssnitch causing erroneous alerts.

[OSD-9349](https://issues.redhat.com/browse/OSD-9349)

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -35,17 +35,15 @@ var fedramp = os.Getenv("FEDRAMP") == "true"
 const (
 	deadMansSnitchAPISecretKey    = "deadmanssnitch-api-key"
 	DeadMansSnitchFinalizerPrefix = "dms.managed.openshift.io/deadmanssnitch-"
-	// These can be removed once Hive is promoted past f73ed3e in all environments
-	// Support for these conditions was removed in https://github.com/openshift/hive/pull/1604
-	legacyHivev1RunningHibernationReason  = "Running"
-	legacyHivev1ResumingHibernationReason = "Resuming"
+	// This can be removed once Hive is promoted past f73ed3e in all environments
+	// Support for this condition was removed in https://github.com/openshift/hive/pull/1604
+	legacyHivev1RunningHibernationReason = "Running"
 )
 
 var validHibernationReasons = []string{
 	hivev1.ResumingOrRunningHibernationReason,
-	// These can be removed once Hive is promoted past f73ed3e in all environments
-	// Support for these conditions was removed in https://github.com/openshift/hive/pull/1604
-	legacyHivev1ResumingHibernationReason,
+	// This can be removed once Hive is promoted past f73ed3e in all environments
+	// Support for this condition was removed in https://github.com/openshift/hive/pull/1604
 	legacyHivev1RunningHibernationReason,
 }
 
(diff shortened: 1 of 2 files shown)


jharrington22:
cc/ @suhanime @2uasimojo

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -617,18 +615,29 @@ func (r *ReconcileDeadmansSnitchIntegration) deleteDMSClusterDeployment(dmsi *de
 }
 
 func instancesAreRunning(cd hivev1.ClusterDeployment) bool {
+	// Get hibernation PowerState a new ClusterDeployment Status field indicating if the cluster is running
+	// ie. The cluster is not "Resuming" if the PowerState is "Running", the cluster is operational.
+	// If the field is blank we move on and check the legacy reasons (It may be blank if the running version of
+	// Hive on cluster doesn't yet support it)
+	if cd.Status.PowerState != "" && cd.Status.PowerState == "Running" {
```
2uasimojo, line 622: The empty string check is redundant.
```suggestion
	if cd.Status.PowerState == "Running" {
```

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -35,17 +35,15 @@ var fedramp = os.Getenv("FEDRAMP") == "true"
...
-	// Support for these conditions was removed in https://github.com/openshift/hive/pull/1604
-	legacyHivev1RunningHibernationReason  = "Running"
-	legacyHivev1ResumingHibernationReason = "Resuming"
+	// This can be removed once Hive is promoted past f73ed3e in all environments
+	// Support for this condition was removed in https://github.com/openshift/hive/pull/1604
+	legacyHivev1RunningHibernationReason = "Running"
 )
 
 var validHibernationReasons = []string{
 	hivev1.ResumingOrRunningHibernationReason,
```
jharrington22, line 44: I probably don't need to check this anymore since PowerState should exist.

2uasimojo, line 44: This should just go away. It leaves situations where you can still "pass" if the cluster is resuming.
```suggestion
```
(Or get rid of `validHibernationReason` entirely and just inline the check for `Running`.)

2uasimojo: Agree, see below.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
suhanime: Checking !=  nil might avoid panics imo

2uasimojo: I don't think the compiler will even let you check for `nil` on a string type.

suhanime: ah right !

jharrington22: Yeap :)

jharrington22:
@2uasimojo @suhanime I had to update the tests to include the new `Status.PowerState` and then added a test for the "legacy" Hibernation condition. I left the const `legacyHivev1RunningHibernationReason` to ensure there is no confusion about whether or not its required once Hive is updated everywehere.

2uasimojo (commented):
The empty string check is still redundant, but that's not a show-stopper.

/lgtm

jharrington22:
> The empty string check is still redundant, but that's not a show-stopper.
> 
> /lgtm

Ah, yeah fixed thanks!

PR #131, opened 2022-02-01
Improve Fedramp Implementation

sam-nguyen7:
Currently the fedramp variable value is not being check, causing log messages reading fedramp=true no matter what. PR ensures that:
- Log message reads fedramp only when fedramp=true
- Only check once at operator start up, instead of every reconcile loop for every clusterDeployment

Tested on stage cluster.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -14,6 +14,12 @@
 
 package config
 
+import (
+	"fmt"
+	"os"
+	"strconv"
+)
+
 const (
 	OperatorName      string = "deadmanssnitch-operator"
 	OperatorNamespace string = "deadmanssnitch-operator"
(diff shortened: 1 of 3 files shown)


sam-nguyen7:
/hold
for testing

config/config.go
```diff
@@ -25,3 +31,31 @@ const (
...
 )
+
+var isFedramp = false
+
+// SetIsFedramp gets the value of fedramp
+func SetIsFedramp() error {
+	// Get fedramp var
+	fedramp, ok := os.LookupEnv("FEDRAMP")
+	if !ok {
+		// If lookup fails, default to false
```
rafael-azevedo, line 42: unnecessary comment

config/config.go
```diff
@@ -25,3 +31,31 @@ const (
 	// if the cluster is OSD (managed) or not
 	ClusterDeploymentManagedLabel string = "api.openshift.com/managed"
 )
+
+var isFedramp = false
+
+// SetIsFedramp gets the value of fedramp
+func SetIsFedramp() error {
+	// Get fedramp var
```
rafael-azevedo, line 39: unnecessary comment

config/config.go
```diff
@@ -25,3 +31,31 @@ const (
...
+	}
+
+	fedrampBool, err := strconv.ParseBool(fedramp)
+	if err != nil {
+		return fmt.Errorf("Invalid value for FedRAMP environment variable. %w", err)
+	}
+
+	isFedramp = fedrampBool
+	if isFedramp {
+		fmt.Println("Running in fedramp environment.")
```
rafael-azevedo, line 53: This function is responsible for setting the variable up not sure you want to just print to console here.

PR #134, opened 2022-03-04
Anonymize dms to use clusterID

sam-nguyen7:
When creating new snitches if in fedRAMP mode anonymize by using the clusterID instead of the clusterName

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -361,17 +361,16 @@ func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1
 		var snitch dmsclient.Snitch
 		if len(snitches) > 0 {
 			snitch = snitches[0]
-		} else {
-			newSnitch := dmsclient.NewSnitch(snitchName, dmsi.Spec.Tags, "15_minute", "basic")
-			newSnitch.Notes = fmt.Sprintf(`cluster_id: %s
+		}
+		newSnitch := dmsclient.NewSnitch(snitchName, dmsi.Spec.Tags, "15_minute", "basic")
+		newSnitch.Notes = fmt.Sprintf(`cluster_id: %s
 runbook: https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md`, snitchName)
-			// add escaping since _ is not being recognized otherwise.
-			newSnitch.Notes = "```" + newSnitch.Notes + "```"
-			logger.Info(fmt.Sprint("Creating snitch:", snitchName))
-			snitch, err = dmsc.Create(newSnitch)
-			if err != nil {
-				return err
-			}
+		// add escaping since _ is not being recognized otherwise.
+		newSnitch.Notes = "```" + newSnitch.Notes + "```"
+		logger.Info(fmt.Sprint("Creating snitch:", snitchName))
+		snitch, err = dmsc.Create(newSnitch)
+		if err != nil {
+			return err
 		}
 
 		ReSnitches, err := dmsc.FindSnitchesByName(snitchName)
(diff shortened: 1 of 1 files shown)


jewzaam:
/hold

Would like to see PD conversations resolved in OSD-9561 and adjust work for DMS (OSD-9562) if needed.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -344,7 +345,9 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
 func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
 	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
 	snitchName := utils.DmsSnitchName(cd.Spec.ClusterName, cd.Spec.BaseDomain, dmsi.Spec.SnitchNamePostFix)
-
+	if config.IsFedramp() {
+		snitchName = getClusterID(cd.Namespace)
```
yithian, line 349: there's no need to make this a function, imo, since it's only used here

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -637,3 +640,8 @@ func getCondition(conditions []hivev1.ClusterDeploymentCondition, t hivev1.Clust
 	}
 	return nil
 }
+
+func getClusterID(clusterNS string) string {
```
yithian, line 644: this can probably be removed since it's only used in one place

yithian (changes requested):

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
rafael-azevedo: @sam-nguyen7 I would keep this function if you pull in functionally to actually grab the data from the API as well. Having just one line of code wrapped in a function is a bit of over engineering.

rafael-azevedo: @sam-nguyen7 I think you also be able to get the clusterID from a label in the Clusterdeployment. That could live in here.

luis-falcon: i like having it in a function. more space to grow. let us say that we want to get/parse the ClusterID in the future from somewhere else or another file. Then we already have a function that does it. I would put it on a utils package though.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -361,7 +364,7 @@ func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1
 		} else {
```
rafael-azevedo, line 364: You don't need the else here please remove it

yithian: unless i'm misreading things, that `else` wasn't part of this commit. should @sam-nguyen7 remove it anyways?

rafael-azevedo: It wasn't but seems to be an easy removal :)

blrm:
@jewzaam Are you ok to drop the hold on this one given the discussion in https://issues.redhat.com/browse/OSD-9561 ?

sam-nguyen7:
/hold
for testing

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -344,7 +344,10 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
 func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
 	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
 	snitchName := utils.DmsSnitchName(cd.Spec.ClusterName, cd.Spec.BaseDomain, dmsi.Spec.SnitchNamePostFix)
-
+	if config.IsFedramp() {
+		clusterID := cd.Labels["id"]
```
rafael-azevedo, line 348: I'd remove this variable completely and add a comment that the snitch uses the clusterID. Otherwise this is bacon

PR #136, opened 2022-03-15
Create DMS secrets on cluster install

supreeth7:
### **What type of PR is this?**
Refactor

### What this PR does / why we need it?
- Addons now wait for both the PagerDuty and DMS secrets to be available in the openshift-monitoring namespace before considering themselves ready. Addons now wait the extra 10-30 minutes after install for CVO to stop transitioning hence this extra delay sets off a timeout set by the Addon's team to trigger the Addon as a failed install.
- This PR ensures that DMS secrets are created on cluster installation without having to wait for ClusterOperators to transition to Ready state.
- This further ensures consistency with PagerDuty operator as pd-secrets are being created on cluster install.
- DMS resources are deleted when a cluster is Hibernating.
- Going by the JIRA, it is agreed that the check for ClusterOperators status for resuming cluster is good idea rather than checking COs status for new cluster installation. PagerDuty Operator handles this gracefully.

### Which Jira/Github issue(s) this PR fixes?
Relates to: [OSD-10039](https://issues.redhat.com/browse/OSD-10039)

### Pre-checks (if applicable):
- [x] Tested the changes against a stage cluster.

### DIFF
--- a/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -224,32 +224,34 @@ func (r *ReconcileDeadmansSnitchIntegration) Reconcile(request reconcile.Request
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		specIsHibernating := clusterdeployment.Spec.PowerState == hivev1.HibernatingClusterPowerState
 
+		// Check if the cluster is hibernating
+		specIsHibernating := clusterdeployment.Spec.PowerState == hivev1.HibernatingClusterPowerState
 		if specIsHibernating {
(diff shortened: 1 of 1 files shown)


jewzaam (changes requested):
> The cluster's PagerDuty service isn't enabled until the ClusterOperators have completely transitioned to Ready state further ensuring there would be no DMS alerts firing for resuming cluster until COs have settled down.

This is not true.  DMS integration to PD is via a single PD service.  It is not using the per-cluster PD service.

I think that's fine though.  As noted in the JIRA, we used to create the DMS secret as soon as CD.spec.installed = true.  So I'll approve.  But please
1. update description to remove the invalid comment above
2. verify DMS setup early is fine (I didn't validate the statement about installed=true)

/approve

PR #137, opened 2022-03-16
Fix clusterID for fedramp

sam-nguyen7:
Snitch note should output internal clusterID for fedramp or external clusterID for standard OSD.


pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -344,9 +344,10 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
 func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
 	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
 	snitchName := utils.DmsSnitchName(cd.Spec.ClusterName, cd.Spec.BaseDomain, dmsi.Spec.SnitchNamePostFix)
+	clusterID := cd.Spec.ClusterMetadata.ClusterID
```
dustman9000, line 347: What if `cd.Spec.ClusterMetadata` is nil?

sam-nguyen7: Good point. Previously (before my PRs) this has never been checked. https://github.com/openshift/deadmanssnitch-operator/blob/9a1a9278f8b40339092d937046935737705d7964/pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go#L364 

If we are checking for `cd.Spec.ClusterMetadata` for nil here, should we also be checking if `cd.Spec` and/or `cd`?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -605,3 +612,44 @@ func (r *ReconcileDeadmansSnitchIntegration) deleteDMSClusterDeployment(dmsi *de
 	return nil
 
 }
+
+func instancesAreRunning(cd hivev1.ClusterDeployment) bool {
```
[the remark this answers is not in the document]
sam-nguyen7: Removed.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -345,15 +347,20 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
...
 func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
 	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	if cd.Spec.ClusterMetadata == nil {
+		return errors.New("ClusterMetadata is nil pointer")
+	}
+
 	snitchName := utils.DmsSnitchName(cd.Spec.ClusterName, cd.Spec.BaseDomain, dmsi.Spec.SnitchNamePostFix)
+	clusterID := cd.Spec.ClusterMetadata.ClusterID
 	if config.IsFedramp() {
-		// Use the clusterID as snitchName
```
cblecker, line 350: This comment went away?

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -345,15 +347,20 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
 // create snitch in deadmanssnitch.com with information retrived from dmsi cr as well as the matching cluster deployment
 func (r *ReconcileDeadmansSnitchIntegration) createSnitch(dmsi *deadmanssnitchv1alpha1.DeadmansSnitchIntegration, cd *hivev1.ClusterDeployment, dmsc dmsclient.Client) error {
 	logger := log.WithValues("DeadMansSnitchIntegration.Namespace", dmsi.Namespace, "DMSI.Name", dmsi.Name, "cluster-deployment.Name:", cd.Name, "cluster-deployment.Namespace:", cd.Namespace)
+	if cd.Spec.ClusterMetadata == nil {
+		return errors.New("ClusterMetadata is nil pointer")
```
cblecker, line 351: This is kind of a weird error. Why is this a nil pointer, and what broke to flat this as an error?

sam-nguyen7: @cblecker logic has been modified to handle this.

pkg/controller/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -345,15 +347,16 @@ func (r *ReconcileDeadmansSnitchIntegration) dmsAddFinalizer(dmsi *deadmanssnitc
...
+	if err != nil {
+		return err
 	}
-	ssName := utils.SecretName(cd.Spec.ClusterName, dmsi.Spec.SnitchNamePostFix)
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: ssName, Namespace: cd.Namespace}, &hivev1.SyncSet{})
+	snitchName := getSnitchName(*cd, dmsi.Spec.SnitchNamePostFix, config.IsFedramp())
 
-	if errors.IsNotFound(err) {
+	ssName := utils.SecretName(cd.Spec.ClusterName, dmsi.Spec.SnitchNamePostFix)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: ssName, Namespace: cd.Namespace}, &hivev1.SyncSet{})
```
rafael-azevedo, line 358: It looks like we are not checking if `err != nil` .

So we get the same behavior for `err == nil` and `err !errors.IsNotFound(err) `

@cblecker Just checking if this is the intended behavior or should we wrap this with an if err != nil

rafael-azevedo: @sam-nguyen7 I'd put the check if err!= nil here before we can merge

PR #143, opened 2022-08-23
OSD-11800 Upgrade to operator-sdk 1.21.1

rendhalver:
[OSD-11800](https://issues.redhat.com//browse/OSD-11800) Upgrade operator-sdk 1.21


main.go
```diff
@@ -0,0 +1,143 @@
...
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "3a94b80a.deadmanssnitch.managed.openshift.io",
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
```
mrbarge, line 105: I note the old operator was performing leader-for-life election (the `Become()` call [here](https://github.com/openshift/deadmanssnitch-operator/blob/09c3c0aae5e866f53a9bf0b0045db13c3b554845/cmd/manager/main.go#L86)) [0], it currently appears to not be doing any leader election at all. In the interests of maintaining existing behaviour we should add this back as we have done for other migrated operators.

[0] https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/

controllers/deadmanssnitchintegration/event_handlers_test.go
```diff
@@ -0,0 +1,182 @@
...
+			obj:              &hivev1.ClusterDeployment{},
+			pdiObjs:          []runtime.Object{},
+			expectedRequests: 0,
+		},
+		{
+			name: "empty ClusterDeployment",
+			obj: &hivev1.ClusterDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"pdiWatching": "clusterDeployment1",
```
mrbarge, line 58: Though it ultimately probably doesn't matter from a testing perspective, renaming this to something that doesn't carry pagerduty-operator connotations might be better. 

```suggestion
						"dmsiWatching": "clusterDeployment1",
```

Same goes for all other `pdi` references.

main.go
rendhalver: Ah thank you for the explanation.
I wasn't quite sure how that worked and thought setting the flag was enough but now it makes more sense.
I will get that updated and push up the fix.

controllers/deadmanssnitchintegration/event_handlers_test.go
rendhalver: Thanks for catching that. I will get those fixed.

main.go
```diff
@@ -0,0 +1,150 @@
...
+	err = leader.Become(context.TODO(), "memcached-operator-lock")
```
mrbarge, line 107: It should use the same lock file name it did previously

rendhalver: Thank you.
I have fixed that.

mrbarge:
@jewzaam @robotmaxtron @rafael-azevedo could we please ask for some :eyes: on this?

Pre-merge testing has been conducted successfully with a report available at OSD-12773. 

The change here is largely similar to PDOperator's [equivalent](https://github.com/openshift/pagerduty-operator/pull/199) which has gone live okay.

rendhalver:
The commits in the are all over the place so I am going to squash the branch to keep the commit history a little cleaner. 
/label tide/merge-method-squash

PR #150, opened 2023-03-06
OSD-15132 Reduce DMS Operator Logs

dakotalongRH:
Set the default logging level to production and move certain logs to the debug level.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -74,7 +75,9 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	opts := zap.Options{
-		Development: true,
+		Development: false,
+		TimeEncoder: zapcore.RFC3339TimeEncoder,
+		StacktraceLevel: zapcore.DPanicLevel,
```
mjlshen, line 80: Weird that CI didn't catch this, but can we format this with a `go fmt ./...`?

mjlshen (commented):
Based on the intent of the story, lgtm, but I'm not very familiar with this codebase so I will leave it up to the OWNERS

PR #158, opened 2023-05-30
OSD-13714 - Improving handling of multi-snitch potential cases

bdematte:
Improving the handling of situation where potentially multiple snitches have been created for the same service (multiple integration with same postfix, potential partial retrieve from DMS service, ...) : 
- initial implementation was potentially trying to validate a snitch which wasn't the one identified as 'Pending' (as using result of 2 different retrieve and DMS doesn't guarantee the ordering)
- in case of multiple snitches, we could potentially never try to activate some of the snitches

Underlying JIRA Card : [OSD-13714](https://issues.redhat.com/browse/OSD-13714)

### DIFF
--- a/controllers/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
+++ b/controllers/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
@@ -330,20 +330,16 @@ func (r *DeadmansSnitchIntegrationReconciler) createSnitch(dmsi *deadmanssnitchv
 				return err
 			}
 
-			var snitch dmsclient.Snitch
 			if len(snitches) <= 0 {
 				newSnitch := dmsclient.NewSnitch(snitchName, dmsi.Spec.Tags, "15_minute", "basic")
 				newSnitch.Notes = fmt.Sprintf("cluster_id: %s\nrunbook: https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md", clusterID)
 				// add escaping since _ is not being recognized otherwise.
 				newSnitch.Notes = "```" + newSnitch.Notes + "```"
 				logger.Info(fmt.Sprint("Creating snitch:", snitchName))
-				snitch, err = dmsc.Create(newSnitch)
+				_, err = dmsc.Create(newSnitch)
 				if err != nil {
 					return err
 				}
-			}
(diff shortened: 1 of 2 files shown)


controllers/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
```diff
@@ -330,21 +330,17 @@ func (r *DeadmansSnitchIntegrationReconciler) createSnitch(dmsi *deadmanssnitchv
...
 
-			var snitch dmsclient.Snitch
 			if len(snitches) <= 0 {
 				newSnitch := dmsclient.NewSnitch(snitchName, dmsi.Spec.Tags, "15_minute", "basic")
 				newSnitch.Notes = fmt.Sprintf("cluster_id: %s\nrunbook: https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md", clusterID)
 				// add escaping since _ is not being recognized otherwise.
 				newSnitch.Notes = "```" + newSnitch.Notes + "```"
 				logger.Info(fmt.Sprint("Creating snitch:", snitchName))
-				snitch, err = dmsc.Create(newSnitch)
+				_, err = dmsc.Create(newSnitch)
```
boranx, line 339: would this cause the snitches to be re-created for restart situations? (like operator-restart/cycling)

boranx (commented):
/lgtm

controllers/deadmanssnitchintegration/deadmanssnitchintegration_controller.go
boranx, line 339: tbh, the whole case is in `k8errors.IsNotFound(err)`,  so you think it could be an issue where the k8s client got a momentary err for all of those existing snitches?

bdematte: I am not convinced this is the error even though it might. Nevertheless, the code as it is doesn't make sense (checking a snitch status to validate a different one potentially) so this deserves to be fixed in any case. 
Without logs or anything, it is difficult to investigate much more, and we didn't have reoccurrence (except maybe with activation/deactivation of CAD handling from PD stats) which rules out behaviour related to promotion/restart and prevent from digging further on recent case to get some logs.

boranx (commented):
/lgtm

PR #149, opened 2023-01-09
Enable watching clusterdeployments with limited-support label

sam-nguyen7:
Revert https://github.com/openshift/deadmanssnitch-operator/commit/9a1a9278f8b40339092d937046935737705d7964 to watch clusterdeployments with limited-support label to allow configuration anomaly detection (CAD) to handle CHGM alerts

for [OSD-12958](https://issues.redhat.com//browse/OSD-12958)

PR #174, opened 2024-01-05
OSD-19710: no snitch for limited support clusters unless supportex

typeid:
This is a followup of the recent refactor done in configuration-anomaly-detection (https://github.com/openshift/configuration-anomaly-detection/pull/246).

With the most recent CAD refactor, we no longer need [the workaround](https://github.com/openshift/deadmanssnitch-operator/pull/149) to handle "resolving" pagerduty alerts, thus don't need to route alerts for clusters that are in limited support through CAD. Therefore, we can disable the snitches for clusters that are in limited support. However, with the recent addition of support exceptions, we still need to handle alerts regardless of the cluster's support state.

This PR disables snitches for clusters that are in limited support **unless they have a support exception**, thus allowing a full transparent flow for CAD. 

Tested scenarios on staging:

- Out of limited support, no supportex =  dms integration
- In limited support, no supportex = no dms integration
- In limited support, supportex (ext-managed.openshift.io/support-exception cd label) = dms integration
- Out of limited support, supportex (ext-managed.openshift.io/support-exception cd label) = dms integration

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /go/src/github.com/openshift/deadmanssnitch-operator
 ENV GOFLAGS=""
 RUN make go-build
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8-1072
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9-1029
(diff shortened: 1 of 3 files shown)


rafael-azevedo (approved):
/lgtm

PR #180, opened 2024-04-19
Upgraded golang/x/net to 0.24 to fix vuln

npecka:
Per https://issues.redhat.com/browse/OSD-22347


mjlshen:
@npecka it's complaining about not having the latest ubi-minimal tag in its `build/Dockerfile`, in practice we need to bug people to get #179 merged (and wait for dependabot to create the next) I can chase down approvers.

PR #218, opened 2025-08-07
Boilerplate: Update to 3b137befb97d00fb0d5d7318ba4c172cde2884cc

tnierman:
https://issues.redhat.com/browse/SREP-811

---

Yet another boilerplate update :tada:, as well as an update to a Konflux issue (see #219)

Boilerplate changes are the result of `make boilerplate-update boilerplate-commit`

---

Conventions:
- openshift/golang-osd-operator: No change ---
https://github.com/openshift/boilerplate/compare/2bf1d94f0c8a6d5adbcf5841ed0a145a41d5ad16...3b137befb97d00fb0d5d7318ba4c172cde2884cc

commit: 4c0b01c6046f85c549f675a1cfad32768cc82926
author: Claudio Busse
Feat: add CAD_PAGERDUTY_ROUTING_KEY to e2e template

commit: 51799b53b00ee90b823f4c13b487a4c5211ba9d3
author: Claudio Busse
Fix: don't overwrite dependabot config in osd-container-image

PR #219, opened 2025-08-08
Fix failing konflux job

tnierman:
https://issues.redhat.com/browse/SREP-811

---

Fixes a failing Konflux job by updating the bundle image.


tnierman:
Seems like there's a catch-22 situation between these changes and the boilerplate update in #218. The easiest workaround here is to bundle these together as a single PR

/close

PR #290, opened 2026-05-19
chore(SREP-4514: Enable code rabbit inheritance for repo)

devppratik:
Enable inheritance from upstream openshift coderabbit repo with exclusion for certain folders.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,10 @@
+inheritance: true
+reviews:
+  path_filters:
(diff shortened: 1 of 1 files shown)


clcollins:
/close
Rebased onto current master in #292 to pick up the boilerplate update from #291, which was causing the `validate` job to fail (stale UBI base image tags). Your commit is authored by you in the new PR — you'll get full author credit. Thanks @devppratik!

